PR #2, opened 2025-06-06
Add minimal devenv

chromy:


NicoHinderling (approved):
great to add the gitignore line we discussed offline before landing but otherwise, 🚢

PR #6, opened 2025-06-09
Set up build and test workflows

trevor-e:
Adds a Makefile to standardize interacting with this repo, and adds Github workflows on top of the Makefile for CI checks.


.flake8
trevor-e, line 1: I don't see this in the main repo but Cursor recommended it. I think eventually sentry is switching to a new tool called `ruff`, but that doesn't seem ready yet.

app-size-analyzer.code-workspace
```diff
@@ -0,0 +1,129 @@
+{
```
NicoHinderling, line 1: should we add this to the repo vs to the .gitignore? unclear to me, but fwiw it's ignored for the monorepo

NicoHinderling (approved):
stamping aside from comments left

app-size-analyzer.code-workspace
trevor-e: Hmm my understanding is this is supposed to be checked in, similar to Xcode's workspace file.

PR #7, opened 2025-06-09
Create initial devservices.yml file

NicoHinderling:
pretty much a copy of the config.yml from snuba, but adapted for this service

### DIFF
--- a/devservices/config.yml
+++ b/devservices/config.yml
@@ -1,0 +1,59 @@
+# Ignored by docker compose, used by devservices
+x-sentry-service-config:
+  version: 0.1
+  service_name: launchpad
+  dependencies:
+    redis:
+      description: Shared instance of redis used by sentry services
+      remote:
+        repo_name: sentry-shared-redis
+        branch: main
+        repo_link: https://github.com/getsentry/sentry-shared-redis.git
+    kafka:
+      description: Shared instance of kafka used by sentry services
+      remote:
+        repo_name: sentry-shared-kafka
+        branch: main
+        repo_link: https://github.com/getsentry/sentry-shared-kafka.git
+    launchpad:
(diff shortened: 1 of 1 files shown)


IanWoodard (commented):
Looks good overall! Found a few things that need to be touched up, but otherwise this is looking pretty solid! Were you able to get it working locally? You'll need to include devservices as a dev dependency so you can run it from your virtual environment.
@hubertdeng123 let me know if you notice anything I missed.

devservices/config.yml
```diff
@@ -0,0 +1,59 @@
...
+  launchpad:
+    image: ghcr.io/getsentry/launchpad:latest
+    ports:
+      - 127.0.0.1:1218:1218
+      - 127.0.0.1:1219:1219
+    command:
+      - devserver
+      - --${LAUNCHPAD_NO_WORKERS:+no-workers}
+    healthcheck:
+      test: curl -f http://localhost:1218/health_envoy
```
hubertdeng123, line 34: Just wanted to confirm that this is the correct endpoint for a healthcheck. This should be changed from localhost:1218 as it will conflict with snuba

devservices/config.yml
```diff
@@ -0,0 +1,59 @@
...
+      interval: 5s
+      timeout: 5s
+      retries: 3
+    environment:
+      PYTHONUNBUFFERED: 1
```
hubertdeng123, line 39: Is this applicable here? I think it is specific to snuba AFAIK

devservices/config.yml
```diff
@@ -0,0 +1,59 @@
...
+      LAUNCHPAD_SETTINGS: docker
```
hubertdeng123, line 40: A lot of these env variables are snuba specific, wondering if they could be removed at least for now until they should be added

PR #10, opened 2025-06-09
Implement binary range mapping

trevor-e:
Starts implementing some of our old `BinarySeparator` class. It doesn't do any advanced analysis for now, mostly just attributes ranges to sections. The other details can come later on.

I added a depdendency on `sortedcontainers` to replace our old red-black tree implementation. I think storing the ranges in an ordered list + binary searching the list should provide similar performance guarantees as before, but we can test that out.


scripts/debug_range_mapping.py
trevor-e, line 1: I've found these small scripts helpful for validating the individual parts so far. Will leave in unless anyone objects.

src/launchpad/analyzers/ios.py
```diff
@@ -280,133 +295,9 @@ def _analyze_binary(self, app_bundle_path: Path, executable_name: str) -> Binary
                 symbols=[],
                 swift_metadata=None,
                 sections={},
+                range_map=None,
             )
 
-    def _extract_architectures(self, binary: lief.Binary) -> List[str]:
```
trevor-e, line 285: Moved

src/launchpad/analyzers/macho_parser.py
```diff
@@ -0,0 +1,324 @@
...
+            )
+
+        except Exception as e:
+            logger.debug(f"Failed to extract Swift metadata: {e}")
+            return None
+
+    def get_header_size(self) -> int:
+        """Get the size of the Mach-O header."""
+        # Mach-O header is typically at the beginning
+        # Size varies by architecture but 32 bytes is common for 64-bit
```
NicoHinderling, line 139: are we eventually going to pass the architecture value here to handle the different cases?

src/launchpad/analyzers/macho_parser.py
```diff
@@ -0,0 +1,324 @@
...
+                if section_size > 0 and file_offset > 0:
+                    sections.append((section_name, file_offset, section_size))
+
+        return sections
+
+    def get_dyld_info(self) -> Optional[Any]:
+        """Get DYLD info if available."""
+        if hasattr(self.binary, "dyld_info"):
+            return self.binary.dyld_info
+        return None
```
NicoHinderling, line 180: stubbed for now? i dont see it being used yet and ideally we avoid the `Any` type, right?

src/launchpad/analyzers/macho_parser.py
```diff
@@ -0,0 +1,324 @@
...
+                section_name = getattr(section, "name", "unknown")
+                file_offset = getattr(section, "offset", 0)
+                section_size = getattr(section, "size", 0)
+
+                if section_size > 0 and file_offset > 0:
+                    sections.append((section_name, file_offset, section_size))
+
+        return sections
+
+    def get_dyld_info(self) -> Optional[Any]:
```
NicoHinderling, line 176: I see both `X | None` and `Optional` used in this file. I think ideally we just use one semantic style. According to the internet, as long as you aren't worried about supporting Python < 3.10, `X | None` is encouraged

src/launchpad/analyzers/macho_parser.py
```diff
@@ -0,0 +1,324 @@
...
+        Returns:
+            (offset, size) tuple or None
+        """
+        if hasattr(self.binary, "code_signature"):
+            code_sig = self.binary.code_signature
+            if hasattr(code_sig, "data_offset") and hasattr(code_sig, "data_size"):
+                return (code_sig.data_offset, code_sig.data_size)
+        return None
+
+    def _cpu_type_to_string(self, cpu_type: int) -> Optional[str]:
```
NicoHinderling, line 194: should we put internal methods like these (I assume that's the indication of the `_` prefix) at the bottom of the file?

src/launchpad/analyzers/macho_parser.py
```diff
@@ -0,0 +1,324 @@
...
+            List of offsets where the pattern was found
+        """
+        try:
+            content = self.get_section_content(section_name)
+            if content is None:
+                return []
+
+            offsets = []
+            start = 0
+            while True:
```
NicoHinderling, line 274: while can be technically fine, `while True` seems like a code smell. Should we maybe add a maximum # of iterations?

src/launchpad/analyzers/macho_parser.py
```diff
@@ -0,0 +1,324 @@
...
+                offsets.append(pos)
+                start = pos + 1
+
+            return offsets
+
+        except Exception as e:
+            logger.debug(f"Failed to search bytes in section {section_name}: {e}")
+            return []
+
+    def extract_strings_from_section(self, section_name: str, min_length: int = 4) -> list[tuple[int, str]]:
```
NicoHinderling, line 287: just curious, how come min_length defaults to 4?

src/launchpad/analyzers/range_mapping_builder.py
```diff
@@ -0,0 +1,485 @@
...
+                        region.start,
+                        region.end,
+                        BinaryTag.UNMAPPED,
+                        f"padding_gap_{region.size}_bytes",
+                        allow_partial=True,  # Use partial splitting for padding gaps
+                    )
+                else:
+                    logger.warning(f"Large unmapped region: {region.size} bytes at offset {region.start}")
+        except Exception as e:
+            logger.debug(f"Failed to map remaining gaps: {e}")
```
NicoHinderling, line 485: honestly didn't look too closely at this file since it's so bespoke and just ripped from our existing code. Will just defer to shadow / manual testing for reliability of this logic

src/launchpad/models/__init__.py
```diff
@@ -47,4 +48,9 @@
     "AndroidAppInfo",
     "AndroidBinaryAnalysis",
     "AndroidAnalysisResults",
+    # Range mapping models
+    "BinaryTag",
+    "Range",
+    "RangeConflict",
+    "RangeMap",
 ]
```
NicoHinderling, line 56: (just my inner monologue) my unfamiliarity with python is shining here, but I'm guessing at some point we'll want to separate some of these into separate groups

src/launchpad/analyzers/macho_parser.py
trevor-e: going to stub this out for now

src/launchpad/analyzers/macho_parser.py
trevor-e: good catch, switched to the union version everywhere

src/launchpad/py.typed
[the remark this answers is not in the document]
trevor-e: I believe this signals that our package is typed. Python uses a lot of these "marker" files.

NicoHinderling (approved):

PR #13, opened 2025-06-10
Remove extra whitespace

chromy:


trevor-e (approved):
Was this triggered by the linter? It would be ideal to figure out why neither CI (currently green on main) nor `make autofix` caught this.

chromy:
I don't think linter checks these, I happened to see these.

PR #16, opened 2025-06-10
[wip] Create basic version of launchpad service

NicoHinderling:
First attempt at creating our service powered by asyncIO and arroyo. Works locally by doing the following:

```
$ devservices up
$ launchpad serve
```
and then in a separate terminal window, I execute `make test-kafka-message` and check the services logs

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,53 @@
+# Use Python 3.12 slim image
+FROM python:3.12-slim-bookworm
+
+# Set environment variables
+ENV PYTHONUNBUFFERED=1 \
+    PYTHONDONTWRITEBYTECODE=1 \
+    PIP_NO_CACHE_DIR=1 \
+    PIP_DISABLE_PIP_VERSION_CHECK=1
+
+# Create app user and group
+RUN groupadd --gid 1000 app && \
+    useradd --uid 1000 --gid app --shell /bin/bash --create-home app
+
+# Install system dependencies
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+        curl \
+        git \
+        build-essential \
+        && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+# Set working directory
+WORKDIR /app
+
+# Copy requirements and install Python dependencies
+COPY requirements.txt .
+RUN pip install --no-cache-dir -r requirements.txt
+
+# Copy source code
+COPY src/ ./src/
+COPY pyproject.toml .
+COPY README.md .
+COPY LICENSE .
+
+RUN pip install -e .
+
+# Change ownership to app user
+RUN chown -R app:app /app
+
+# Switch to app user
+USER app
+
+# Expose ports
+EXPOSE 2218
+
+# Health check
+HEALTHCHECK --interval=30s --timeout=5s --start-period=30s --retries=3 \
+    CMD curl -f http://localhost:2218/health || exit 1
+
+# Default command
+CMD ["launchpad", "serve"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,9 +45,5 @@ USER app
 # Expose ports
 EXPOSE 2218
 
-# Health check
-HEALTHCHECK --interval=30s --timeout=5s --start-period=30s --retries=3 \
-    CMD curl -f http://localhost:2218/health || exit 1
-
 # Default command
 CMD ["launchpad", "serve"]
--- a/devservices/config.yml
+++ b/devservices/config.yml
@@ -39,7 +39,7 @@ services:
       KAFKA_TOPICS: launchpad-events
       LAUNCHPAD_HOST: "0.0.0.0" # Inside container, but port binding restricts access
       LAUNCHPAD_PORT: "2218"
-    platform: amd64
+    platform: linux/amd64
     extra_hosts:
       host.docker.internal: host-gateway
     networks:
--- a/devservices/config.yml
+++ b/devservices/config.yml
@@ -39,7 +39,7 @@ services:
       KAFKA_TOPICS: launchpad-events
       LAUNCHPAD_HOST: "0.0.0.0" # Inside container, but port binding restricts access
       LAUNCHPAD_PORT: "2218"
-    platform: linux/amd64
+    platform: amd64
     extra_hosts:
       host.docker.internal: host-gateway
     networks:
--- a/devservices/config.yml
+++ b/devservices/config.yml
@@ -1,0 +1,54 @@
+# Ignored by docker compose, used by devservices
+x-sentry-service-config:
+  version: 0.1
+  service_name: launchpad
+  dependencies:
+    kafka:
+      description: Shared instance of kafka used by sentry services
+      remote:
+        repo_name: sentry-shared-kafka
+        branch: main
+        repo_link: https://github.com/getsentry/sentry-shared-kafka.git
+    launchpad:
+      description: Service that powers preprod artifact analysis
+  modes:
+    default: [kafka]
+    containerized: [kafka, launchpad]
+
+services:
+  launchpad:
+    build:
+      context: ..
+      dockerfile: Dockerfile
+    ports:
+      - 127.0.0.1:2218:2218 # Bind to localhost only - no external access
+    command:
+      - launchpad
+      - serve
+      - --verbose
+    healthcheck:
+      test: curl -f http://localhost:2218/health || exit 1
+      interval: 10s
+      timeout: 5s
+      retries: 3
+      start_period: 30s
+    environment:
+      PYTHONUNBUFFERED: 1
+      KAFKA_BOOTSTRAP_SERVERS: kafka:9093
+      KAFKA_GROUP_ID: launchpad-devservices
+      KAFKA_TOPICS: launchpad-events
+      LAUNCHPAD_HOST: "0.0.0.0" # Inside container, but port binding restricts access
+      LAUNCHPAD_PORT: "2218"
+    platform: linux/amd64
+    extra_hosts:
+      host.docker.internal: host-gateway
+    networks:
+      - devservices
+    labels:
+      - orchestrator=devservices
+    restart: unless-stopped
+
+networks:
+  devservices:
+    name: devservices
+    external: true
--- a/src/launchpad/cli.py
+++ b/src/launchpad/cli.py
@@ -2,8 +2,10 @@
 
 from __future__ import annotations
 
+import asyncio
 import json
-import pathlib
+import logging
+import os
 import time
 from pathlib import Path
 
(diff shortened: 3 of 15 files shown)


LAUNCHPAD_SETUP_AND_TESTING.md
```diff
@@ -0,0 +1,419 @@
+# Launchpad Setup and Testing Guide
```
NicoHinderling, line 1: I plan to remove this file and merge some of these details into the README

src/launchpad/cli.py
```diff
@@ -35,18 +38,18 @@ def cli(ctx: click.Context, version: bool) -> None:
 
 
 @cli.command()
-@click.argument("input_path", type=click.Path(exists=True, path_type=pathlib.Path), metavar="INPUT_PATH")
+@click.argument("input_path", type=click.Path(exists=True), metavar="INPUT_PATH")
```
NicoHinderling, line 41: these changes up until the new `devserver` service were requested by precommit

docker-compose.yml
```diff
@@ -0,0 +1,47 @@
...
+    volumes:
+      # Mount source code for live reloading during development
+      - ./src:/app/src
+      - ./tests:/app/tests
+    command:
+      - launchpad
+      - devserver
+      - --verbose
+
+  # Optional local Kafka for standalone development (when not using devservices)
```
hubertdeng123, line 20: Ideally we wouldn't need a separate docker compose file here, it would be integrated with devservices. If there is a need, we'd like to understand why to accomodate. What is the reason why there is a separate kafka/zookeeper instance here? Is zookeeper needed for dev since kafka has kRaft mode?

Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+RUN chown -R app:app /app
+
+# Switch to app user
+USER app
+
+# Expose ports
+EXPOSE 2218
+
+# Health check
+HEALTHCHECK --interval=30s --timeout=5s --start-period=30s --retries=3 \
```
hubertdeng123, line 50: I believe if the healthcheck is defined in the dockerfile, it does not need to be in the docker compsoe file or devservices config.yml file

Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+    CMD curl -f http://localhost:2218/health || exit 1
+
+# Default command
+CMD ["launchpad", "devserver"]
```
hubertdeng123, line 54: Just to confirm - is this a dockerfile that is used to build production images? If so it should likely not be using a devserver but a more production ready server

docker-compose.yml
hubertdeng123: Devservices does not support building/rebuilding images which I see is done here, what kind of benefits does building an image here provide towards development versus running the devserver directly from source? Asking because generally we have avoided doing that to provide a better interactive debugging experience for our devs

docker-compose.yml
```diff
@@ -0,0 +1,47 @@
...
+  kafka:
+    image: confluentinc/cp-kafka:7.4.0
+    hostname: kafka
+    container_name: launchpad-kafka
+    profiles: ["standalone"]
+    depends_on:
+      - zookeeper
+    ports:
+      - "9092:9092"
```
hubertdeng123, line 29: Probably should limit port exposure to just localhost here and below to lower risk exposure

Dockerfile
NicoHinderling: is there a preference of which location? If not, I'll defer to just keep the one in `devservices/config.yml`

Dockerfile
NicoHinderling: I was under the assumption that for production we ideally use the production image from the docker file, but frankly I'm not sure if that's the case. WDYT?

> If so it should likely not be using a devserver but a more production ready server
heh, i just copied snuba's "devserver" naming, despite it originally being "server". I don't think I have any difference in config between dev and prod yet

docker-compose.yml
NicoHinderling: > Ideally we wouldn't need a separate docker compose file here, it would be integrated with devservices. If there is a need, we'd like to understand why to accomodate.

Separate docker-compose file vs devservices
	- I noticed we had this in snuba, so I assumed that having the standalone docker support would be necessary for production. Is just devservices alone sufficient for a prod service? In which case, do we even need the docker-compose.yml file at all in our service?
	
> What is the reason why there is a separate kafka/zookeeper instance here? Is zookeeper needed for dev since kafka has kRaft mode?

The separate instance was just for dev testing the standalone approach. I wasn't familiar with kRaft mode though. Just tested the standalone approach with zookeeper removed and seems to work so pushed that!

> Devservices does not support building/rebuilding images which I see is done here, what kind of benefits does building an image here provide towards development versus running the devserver directly from source? Asking because generally we have avoided doing that to provide a better interactive debugging experience for our devs

It's not for development purposes but rather purely for the production case

it would be great to chat tomorrow so I can better understand the dev vs prod use cases :D

Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+        git \
+        build-essential \
+        && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+# Set working directory
+WORKDIR /app
+
+# Copy requirements and install Python dependencies
```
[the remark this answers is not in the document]
NicoHinderling: meant to use prod deps, good catch

devservices/config.yml
```diff
@@ -0,0 +1,54 @@
...
+      retries: 3
+      start_period: 30s
+    environment:
+      PYTHONUNBUFFERED: 1
+      KAFKA_BOOTSTRAP_SERVERS: kafka:9093
+      KAFKA_GROUP_ID: launchpad-devservices
+      KAFKA_TOPICS: launchpad-events
+      LAUNCHPAD_HOST: "0.0.0.0" # Inside container, but port binding restricts access
+      LAUNCHPAD_PORT: "2218"
+    platform: linux/amd64
```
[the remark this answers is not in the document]
NicoHinderling: looks like both snuba and taskbroker use `linux/amd64`. @hubertdeng123 do you know?

cc @markstory in case you looked into it while standing taskbrokers up

src/launchpad/kafka.py
```diff
@@ -0,0 +1,319 @@
...
+        self,
+        topics: List[str],
+        group_id: str,
+        bootstrap_servers: Optional[str] = None,
+        message_handler: Optional[Callable[[LaunchpadMessage], Any]] = None,
+    ) -> None:
+        self.topics = topics
+        self.group_id = group_id
+        # Ensure bootstrap_servers is always a string
+        self.bootstrap_servers: str = bootstrap_servers or os.getenv("KAFKA_BOOTSTRAP_SERVERS") or "localhost:9092"
```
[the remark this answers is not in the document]
NicoHinderling: good call, yeah i'll throw an exception

src/launchpad/service.py
```diff
@@ -0,0 +1,202 @@
...
+            payload = message.get_json_payload()
+            if not payload:
+                logger.error("Failed to parse message payload")
+                return
+
+            # TODO: Route message to appropriate handler based on message type
+            message_type = payload.get("type", "unknown")
+
+            if message_type == "analyze_ios":
+                self._handle_ios_analysis_sync(payload)
```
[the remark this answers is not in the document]
NicoHinderling: yeah this is very important. I'll add an extra comment as a reminder

Dockerfile
NicoHinderling: synced offline, keeping the dockerfile around precisely for the prod use case

Dockerfile
NicoHinderling: synced offline, removed it from Dockerfile

devservices/config.yml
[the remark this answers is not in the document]
hubertdeng123: Most of our services all use linux/amd64. There has been talk around eventually building arm64 images and supporting that, but last time we checked there are a smaller pool of those available which will result in queuing of jobs in the sentry repo with our test shards. So, for those reasons I'd lean towards `linux/amd64` here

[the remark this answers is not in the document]
markstory: For taskbrokers, the arm builds were added to support self-hosted being arm compatible. At some point in the future, we may start operating taskbrokers on arm hardware, but there are no short term plans for that.

src/launchpad/kafka.py
```diff
@@ -0,0 +1,327 @@
...
+    def run(self) -> None:
+        """Run the Kafka consumer (blocking, like Snuba)."""
+        # Ensure required topics exist before starting consumer
+        try:
+            ensure_topics_exist(
+                bootstrap_servers=self.bootstrap_servers,
+                topic_names=self.topics,
+                num_partitions=1,
+                replication_factor=1,
+            )
```
markstory, line 233: When you get closer to production, you'll need a way to disable this logic as topics, partitions and kafka clusters are configured through getsentry/ops. The streaming teams can help with getting your topic setup.

src/launchpad/service.py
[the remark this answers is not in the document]
markstory: Doing long running IO inside a kafka consumer can get tricky as Kafka requires at least a heartbeat/poll from consumers every 30s or the consumer is considered 'dead' and the topic is rebalanced.

src/launchpad/service.py
```diff
@@ -0,0 +1,204 @@
...
+        self.server = LaunchpadServer(host=server_config["host"], port=server_config["port"])
+
+        # Setup Kafka consumer
+        kafka_config = get_kafka_config()
+        self.kafka_consumer = KafkaConsumer(
+            topics=kafka_config["topics"],
+            group_id=kafka_config["group_id"],
+            bootstrap_servers=kafka_config["bootstrap_servers"],
+            message_handler=self.handle_kafka_message,
+        )
```
markstory, line 39: Will you be able to operate the web and consumer components independently? Being able to do that can help with handling production scale as we can separate the web and consumers into separate k8s deployments.

NicoHinderling: The web server is just for the health check, so it doesnt make sense to separate it. Any scaling needs will solely be around the consumer logic fortunately!

src/launchpad/kafka.py
```diff
@@ -0,0 +1,327 @@
...
+from arroyo.processing.processor import StreamProcessor
+from arroyo.processing.strategies import ProcessingStrategy, ProcessingStrategyFactory
+from arroyo.types import BrokerValue, Commit, Partition
+from confluent_kafka import KafkaError, KafkaException
+from confluent_kafka.admin import AdminClient, NewTopic
+
+logger = logging.getLogger(__name__)
+
+
+def ensure_topics_exist(
```
bmckerry, line 24: Kafka consumers should not be creating the topics they consume from. Topics have to be registered in the `sentry_kafka_schemas` repo and then deployed to production, the docs for the process can be found here: https://github.com/getsentry/ops/blob/master/shared_config/kafka/README.md

Even if you're not deploying to prod yet I'd recommend at least adding the topic to `sentry_kafka_schemas` as that's how the sentry devserver knows which topics to create on your local kafka

fpacifici: > Kafka consumers should not be creating the topics

Replace with 

> Kafka consumers must not be creating the topics

PR #24, opened 2025-06-12
Add ZippedApk artifact

rbro112:
Allows us to parse uploads from the Sentry CLI (normalized to be zipped).


src/launchpad/analyzers/android.py
```diff
@@ -1,14 +1,18 @@
...
+from ..artifacts import APK, AndroidArtifact, ZippedAPK
 from ..models.android import AndroidAnalysisResults, AndroidAppInfo
 
 
 class AndroidAnalyzer:
     """Analyzer for Android apps (.apk, .aab files)."""
 
-    def __init__(self, artifact: APK) -> None:
-        self.artifact = artifact
+    def __init__(self, path: Path) -> None:
```
rbro112, line 14: No super strong opinion yet if we should take the path or the artifact impl. Maybe when we move to running this as a "service" and non-CLI we can take the artifact but doesn't really matter too much to me.

PR #28, opened 2025-06-13
Add treemap breakdown for binary sections

trevor-e:
Now replaces binary treemap elements with a more detailed breakdown of sections. Future work can then break down sections further to get stuff like Swift/ObjC types.

Screenshot 2025-06-13 at 5 35 13 PM

### DIFF
--- a/scripts/debug_range_mapping.py
+++ b/scripts/debug_range_mapping.py
@@ -27,7 +27,7 @@ def analyze_range_mapping(app_path: Path) -> None:
     print(f"Analyzing: {app_path}")
     print("=" * 50)
 
-    analyzer = IOSAnalyzer(enable_range_mapping=True)
+    analyzer = IOSAnalyzer(skip_range_mapping=False)
     artifact = ArtifactFactory.from_path(app_path)
     results = analyzer.analyze(cast(IOSArtifact, artifact))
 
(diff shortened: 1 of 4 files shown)


src/launchpad/analyzers/ios.py
```diff
@@ -57,53 +55,70 @@ def __init__(
...
+            treemap_builder = TreemapBuilder(
+                app_name=app_info.name,
+                platform="ios",
+                download_compression_ratio=0.8,  # iOS apps typically compress to ~80%
+            )
+            treemap = treemap_builder.build_file_treemap(file_analysis)
 
-        return IOSAnalysisResults(
+            # If range mapping is enabled, add binary section details to treemap
+            if self.enable_range_mapping:
```
trevor-e, line 91: I might consolidate some of these flags later on since it's already getting a bit confusing, but it's nice having size features toggleable

src/launchpad/analyzers/ios.py
```diff
@@ -82,28 +67,54 @@ def analyze(self, artifact: IOSArtifact) -> IOSAnalysisResults:
...
-            # Analyze the main executable binary
-        binary_analysis = self._analyze_binary(artifact)
-        logger.info(f"Binary analysis complete, " f"executable size: {binary_analysis.executable_size} bytes")
+        if self.enable_treemap:
+            # Collect all binaries first if range mapping is enabled
+            if self.enable_range_mapping:
+                # Analyze main executable
+                main_executable = artifact.get_plist().get("CFBundleExecutable", "Unknown")
+                app_bundle_path = artifact.get_app_bundle_path()
+                main_binary_path = app_bundle_path / main_executable
```
NicoHinderling, line 83: this syntax works? surprised you dont have to do string interpolation

PR #35, opened 2025-06-16
Add bundletool utils and split APKs from AAB in prep for treemap gen

rbro112:
In prep for Android treemap generation we'll need to first split the APKs from any AAB. This sets us up to do that. Stacked PRs will handle the basic treemap impls.

### DIFF
--- a/src/launchpad/analyzers/android.py
+++ b/src/launchpad/analyzers/android.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
-from launchpad.models.common import FileAnalysis
-
-from ..artifacts import AndroidArtifact
+from ..artifacts import AAB, APK, AndroidArtifact, ZippedAAB, ZippedAPK
 from ..models.android import AndroidAnalysisResults, AndroidAppInfo
+from ..models.common import FileAnalysis
+from ..utils.logging import get_logger
+
+logger = get_logger(__name__)
 
 
 class AndroidAnalyzer:
@@ -18,6 +20,23 @@ class AndroidAnalyzer:
             build=manifest_dict["version_code"] or "Unknown",
             package_name=manifest_dict["package_name"],
         )
+
+        apks: list[APK] = []
+        # Split AAB into APKs, or use the APK directly
+        if isinstance(artifact, AAB):
+            apks = artifact.get_primary_apks()
+        elif isinstance(artifact, ZippedAAB):
+            apks = artifact.get_primary_apks()
+        elif isinstance(artifact, ZippedAPK):
+            apks.append(artifact.get_primary_apk())
+        elif isinstance(artifact, APK):
+            apks.append(artifact)
+        else:
+            raise ValueError(f"Unsupported artifact type: {type(artifact)}")
+
+        logger.debug("Found %d APKs", len(apks))
+
+        # TODO: Implement treemap generation from APKs
 
         return AndroidAnalysisResults(
             app_info=app_info,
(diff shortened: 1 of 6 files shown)


rbro112:
* **#35** <a href="https://app.graphite.dev/github/pr/getsentry/launchpad/35?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/getsentry/launchpad/35?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/launchpad/analyzers/android.py
```diff
@@ -19,6 +21,23 @@ def analyze(self, artifact: AndroidArtifact) -> AndroidAnalysisResults:
             package_name=manifest_dict["package_name"],
         )
 
+        apks: list[APK] = []
```
trevor-e, line 24: Can this be pushed into artifacts? From the analyzer POV I don't see why it would concern itself with these details other than calling `artifact.get_primary_apks()`.

src/launchpad/utils/android/bundletool.py
```diff
@@ -0,0 +1,183 @@
...
+            try:
+                result = subprocess.run(
+                    ["which", "bundletool"],
+                    capture_output=True,
+                    text=True,
+                    check=True,
+                )
+                bundletool_path = result.stdout.strip()
+            except subprocess.CalledProcessError as e:
+                raise FileNotFoundError("bundletool not found in PATH. Install with `brew install bundletool`") from e
```
trevor-e, line 52: note: we'll have to revisit this once we get further along with the Docker stuff

src/launchpad/utils/android/bundletool.py
```diff
@@ -0,0 +1,183 @@
...
+
+            # Extract APKs for the specified device
+            extract_command = [
+                "extract-apks",
+                f"--apks={temp_apks_path}",
+                f"--output-dir={output_dir}",
+                f"--device-spec={temp_device_spec_path}",
+            ]
+            self._run_command(extract_command)
+        finally:
```
trevor-e, line 135: Is this `finally` needed if everything is nested under the original `with` and `delete=True`? Not totally sure how python handles these blocks with each other.

src/launchpad/utils/android/bundletool.py
```diff
@@ -0,0 +1,183 @@
...
+        except subprocess.SubprocessError as e:
+            raise BundletoolError(
+                f"Failed to run bundletool command: {e}",
+                -1,
+                "",
+                str(e),
+            ) from e
+
+        if result.returncode != 0:
+            raise BundletoolError(
```
trevor-e, line 91: I don't think these exceptions are getting logged anywhere, which we probably want to happen somewhere. E.g. in the AAB it just triggers this cleanup and doesn't log anything:
```
        finally:
            cleanup_directory(apks_dir)
```

trevor-e (approved):

src/launchpad/utils/android/bundletool.py
rbro112: I think you're correct here, fixed

src/launchpad/analyzers/android.py
rbro112: Chatted offline, kicking this to future cleanup

src/launchpad/utils/android/bundletool.py
trevor-e: Yep ignore me here 🤦

mclearyredoak: Fixed in the same commit — `--diff-filter=A` now restricts the first command to added files only, matching the bullet point wording.

PR #37, opened 2025-06-16
Fix nested binary breakdowns for plugins/extensions

trevor-e:
Now breaks down nested binaries like those in the app extensions folder. Need to find an app with a watch app to test that case but should be easy to support if it doesn't already work.

Screenshot 2025-06-16 at 6 31 29 PM

### DIFF
--- a/src/launchpad/analyzers/ios.py
+++ b/src/launchpad/analyzers/ios.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import os
+import subprocess
 import time
 from collections import defaultdict
 from pathlib import Path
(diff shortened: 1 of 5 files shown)


src/launchpad/analyzers/ios.py
```diff
@@ -139,6 +135,43 @@ def _extract_app_info(self, xcarchive: ZippedXCArchive) -> IOSAppInfo:
             sdk_version=plist.get("DTSDKName"),
         )
 
+    def _detect_file_type(self, file_path: Path) -> str:
```
trevor-e, line 138: Some files like Mach-O binaries don't have an explicit file extension. In these cases I added a fallback to use the `file` command and determine the type. I have some future cases in mind that depend on this too.

NicoHinderling (approved):

PR #39, opened 2025-06-17
feat: Support provisioning profile data in app info

noahsmartin:
Add provisioning profile info used for app install

### DIFF
--- a/src/launchpad/analyzers/apple.py
+++ b/src/launchpad/analyzers/apple.py
@@ -7,14 +7,21 @@ import subprocess
 import time
 from collections import defaultdict
 from pathlib import Path
-from typing import Dict, List
+from typing import Any, Dict, List, Tuple
 
 import lief
 
 from launchpad.artifacts.apple.zipped_xcarchive import ZippedXCArchive
 from launchpad.artifacts.artifact import AppleArtifact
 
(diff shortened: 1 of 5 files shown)


src/launchpad/analyzers/apple.py
```diff
@@ -66,7 +81,11 @@ def analyze(self, artifact: AppleArtifact) -> AppleAnalysisResults:
         analysis_start_time = time.time()
 
         # Extract basic app information
-        app_info = self._extract_app_info(artifact)
+        if not self.app_info:
```
trevor-e, line 84: Can this automatically call it if not already done? Will simplify tests and other stuff.

src/launchpad/analyzers/apple.py
```diff
@@ -50,6 +57,14 @@ def __init__(
         self.skip_symbols = skip_symbols
         self.skip_range_mapping = skip_range_mapping
         self.skip_treemap = skip_treemap
+        self.app_info: AppleAppInfo | None = None
+
+    def preprocessing(self, artifact: AppleArtifact) -> AppleAppInfo:
```
trevor-e, line 62: `preprocessing` -> `preprocess`

PR #43, opened 2025-06-17
Add initial iOS file duplication insight + insight framework

trevor-e:


src/launchpad/analyzers/apple.py
```diff
@@ -120,12 +124,28 @@ def analyze(self, artifact: AppleArtifact) -> AppleAnalysisResults:
...
+            temp_results = AppleAnalysisResults(
+                app_info=app_info,
+                file_analysis=file_analysis,
+                binary_analysis=binary_analysis,
+                analysis_duration=0,
+                treemap=treemap,
+                insights=InsightResults(),
+            )
+            insights = {
+                "duplicate_files": self.INSIGHTS[0].generate_insight(temp_results),
```
trevor-e, line 139: This part is weird but i'll clean it up when I add in some more insights.

src/launchpad/models/common.py
```diff
@@ -44,65 +51,79 @@ class SymbolInfo(BaseModel):
...
-        data = self.model_dump()
-        data["generated_at"] = self.generated_at.isoformat()
-        return data
+    total_savings: int = Field(..., ge=0, description="Total potential savings in bytes")
+
+
+class DuplicateFilesInsightResult(BaseInsightResult):
+    """Results from duplicate files analysis."""
+
+    has_duplicates: bool = Field(..., description="Whether any duplicate files were found")
```
rbro112, line 117: Wouldn't just the presence of any `duplicate_groups` (non-empty) cover us?

src/launchpad/models/common.py
```diff
@@ -44,65 +51,79 @@ class SymbolInfo(BaseModel):
...
 
+    @property
+    def duplicate_count(self) -> int:
+        """Number of duplicate files (excluding the original)."""
+        return len(self.files) - 1
 
-class BaseAnalysisResults(BaseModel):
-    """Base analysis results structure."""
+
+class BaseInsightResult(BaseModel):
```
rbro112, line 106: Might be good to move these into a `/models/insights/common.py` folder as I imagine there's going to be a lot of insights and they might bloat this file. Don't feel too strongly here but I envision this file getting pretty large if we dump all insights here (and the platform-specific results files too).

src/launchpad/utils/treemap_builder.py
```diff
@@ -405,17 +401,15 @@ def _calculate_category_breakdown(self, file_analysis: FileAnalysis) -> Dict[str
...
-                # Use filesystem block-aligned size for install calculations
-                install_size = self._calculate_aligned_install_size(file_info)
-                download_size = int(install_size * self.download_compression_ratio)
+            # Use filesystem block-aligned size for install calculations
+            install_size = self._calculate_aligned_install_size(file_info)
+            download_size = int(install_size * self.download_compression_ratio)
 
-                breakdown[category]["install"] += install_size
-                breakdown[category]["download"] += download_size
+            breakdown[treemap_type]["install"] += install_size
```
rbro112, line 411: Assuming this is also intentional? Feels a bit unrelated to this change, maybe better to move into a stacked PR?

src/launchpad/models/common.py
rbro112, line 106: From an organizational standpoint, feels like we should move these models into a `models/insights/common.py` file. Then from there we can add more insights/platform-specific ones. Seems this will bloat the size of these result models

rbro112: `InsightResults` can stay here, but IMO should import models from `/models/insights/common.py`

src/launchpad/analyzers/apple.py
```diff
@@ -271,6 +288,12 @@ def _analyze_files(self, xcarchive: ZippedXCArchive) -> FileAnalysis:
             # Calculate hash for duplicate detection
             file_hash = calculate_file_hash(file_path, algorithm="md5")
 
+            # Analyze image if applicable
+            # TODO: image analysis
+            # image_analysis_result = None
+            # if file_type.lower() in {"png", "jpg", "jpeg", "webp"}:
```
rbro112, line 294: Remove this?

src/launchpad/analyzers/apple.py
rbro112: Yeah generally I'm a bit confused here why this block is taking place to generate the temp results, this whole block is a bit hard to follow.

src/launchpad/models/common.py
```diff
@@ -44,65 +51,79 @@ class SymbolInfo(BaseModel):
...
+    path: str = Field(..., description="Relative path in the bundle")
+    size: int = Field(..., ge=0, description="File size in bytes")
+    file_type: str = Field(..., description="File type/extension")
+    hash_md5: str = Field(..., description="MD5 hash of file contents")
+    treemap_type: str = Field(..., description="Type for treemap visualization")
 
 
-class BaseBinaryAnalysis(BaseModel):
-    """Base binary analysis that applies across platforms."""
+class DuplicateFileGroup(BaseModel):
```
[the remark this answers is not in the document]
trevor-e: ugh this was Cursor adding this back

src/launchpad/models/common.py
```diff
@@ -7,33 +7,40 @@
 
 from pydantic import BaseModel, ConfigDict, Field
 
-from .treemap import TreemapType
 
-
-class FileInfo(BaseModel):
-    """Information about a single file."""
+class BaseAnalysisResults(BaseModel):
```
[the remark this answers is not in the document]
trevor-e: sorry forgot to flag, yea the ordering was a bit weird in this file so I changed it which made the diff a bit terrible

src/launchpad/insights/common.py
```diff
@@ -0,0 +1,80 @@
...
+from launchpad.models.common import DuplicateFilesInsightResult, FileAnalysis, FileInfo
+from launchpad.models.treemap import TreemapResults
+
+from ..models.apple import AppleAppInfo, MachOBinaryAnalysis
+
+T_co = TypeVar("T_co", covariant=True)
+
+
+@dataclass
+class InsightsInput:
```
trevor-e, line 18: Created a new data class to pass in as an input rather than the weirdness of before with trying to re-use the results model. That model will stay strictly for the final serialization.

src/launchpad/analyzers/apple.py
```diff
@@ -120,12 +120,26 @@ def analyze(self, artifact: AppleArtifact) -> AppleAnalysisResults:
...
+        if not self.skip_insights:
+            logger.info("Generating insights from analysis results")
+            insights_input = InsightsInput(
+                app_info=app_info,
+                file_analysis=file_analysis,
+                binary_analysis=binary_analysis,
+                treemap=treemap,
+            )
+            insights = AppleInsightResults(
+                duplicate_files=DuplicateFilesInsight().__call__(insights_input),
```
trevor-e, line 133: Got rid of the prior "fanciness" that was hoping to execute the insights in a loop. Now simply sets a value.

PR #36, opened 2025-06-16
Add basic treemap generation for Android

rbro112:
Adds treemap generation for Android. Also makes it so Android is unzipping the underlying APKs and processing the files like iOS (per offline discussion with Trevor). 

Screenshot 2025-06-17 at 10 50 16 PM

PR #46, opened 2025-06-18
Add random keystore signing to split APKs

rbro112:
Bundletool will auto sign APKs with a default keystore if present. On local, my keystore is globally set as I have the Android SDK setup. This adds a small bit of overhead to the file size and total files.

In CI, there’s no Android SDK and therefore no keystore, so the APKs are not signed. This means there’s less files in the generated APKs and less app size. Resulted in my local tests not matching CI in #36.

This will dynamically generate a random keystore and use that to sign (what we did previously at Emerge). Will also need that logic for build distro eventually for the universal APKs to be signed with a rotating key so it’ll be reusable in the future.

### DIFF
--- a/src/launchpad/utils/android/bundletool.py
+++ b/src/launchpad/utils/android/bundletool.py
@@ -3,6 +3,8 @@
 from __future__ import annotations
 
 import json
+import secrets
+import string
 import subprocess
 import tempfile
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


src/launchpad/utils/android/bundletool.py
```diff
@@ -98,40 +100,117 @@ def _run_command(self, command: list[str], **kwargs: Any) -> tuple[int, str, str
...
-
-        # Extract APKs for the specified device
-        extract_command = [
-            "extract-apks",
-            f"--apks={temp_apks_path}",
-            f"--output-dir={output_dir}",
-            f"--device-spec={temp_device_spec_path}",
-        ]
-        self._run_command(extract_command)
+            temp_file.flush()  # Ensure data is written to disk
```
rbro112, line 202: Fix necessary for executing bundletool, turns out things were broken! #36 adds an integration test so we'll be safe moving forward

trevor-e (approved):

PR #50, opened 2025-06-18
feat: Add code signature validation

noahsmartin:


src/launchpad/analyzers/apple.py
```diff
@@ -149,6 +150,19 @@ def _extract_app_info(self, xcarchive: ZippedXCArchive) -> AppleAppInfo:
 
         supported_platforms = plist.get("CFBundleSupportedPlatforms", [])
         is_simulator = "iphonesimulator" in supported_platforms or plist.get("DTPlatformName") == "iphonesimulator"
+
+        # Validate code signature
+        is_valid = True
```
trevor-e, line 155: Should we be defensive and start this with `is_valid = False`?

src/launchpad/models/apple.py
```diff
@@ -37,6 +37,8 @@ class AppleAppInfo(BaseAppInfo):
         None, description="Type of codesigning used (development, adhoc, appstore, enterprise)"
     )
     profile_name: str | None = Field(None, description="Name of the provisioning profile used")
+    is_valid: bool = Field(True, description="Whether the app's code signature is valid")
```
trevor-e, line 40: Can this be `is_code_signature_valid`

src/launchpad/utils/code_signature_validator.py
```diff
@@ -0,0 +1,382 @@
...
+
+        return True
+
+    def _validate_certificates(self, certificates: List[bytes]) -> bool:
+        """Validate certificates."""
+        if not certificates:
+            logger.warning("[Codesign] No certificates found")
+            return False
+
+        # For now, implement a simplified certificate validation
```
noahsmartin, line 368: I need to follow up with the full implementation here

src/launchpad/utils/code_signature_validator.py
```diff
@@ -0,0 +1,382 @@
...
+    def _validate_executable(self) -> BinaryCheckResult:
+        """Validate the executable."""
+        executable = self.archive.get_binary_path()
+        if not executable:
+            raise ValueError("No executable found")
+        fat_binary = lief.MachO.parse(str(executable))
+        if fat_binary is None:
+            raise ValueError("Failed to parse binary")
+
+        self.macho_parser = MachOParser(fat_binary.at(0))
```
trevor-e, line 136: Is this always the first fat binary?

src/launchpad/analyzers/apple.py
```diff
@@ -149,6 +150,19 @@ def _extract_app_info(self, xcarchive: ZippedXCArchive) -> AppleAppInfo:
...
+        errors: List[str] = []
+        try:
+            validator = CodeSignatureValidator(xcarchive)
+            is_valid, errors = validator.validate()
+        except Exception as e:
+            logger.warning(f"Failed to validate code signature: {e}")
+            is_valid = False
+            errors = [str(e)]
+
+        logger.info(f"Code signature validation errors: {errors}")
```
trevor-e, line 165: Can this be moved into the error block? Don't think we need to see its output for the happy path.

src/launchpad/models/apple.py
```diff
@@ -37,6 +37,8 @@ class AppleAppInfo(BaseAppInfo):
...
+    errors: List[str] = Field(default_factory=list, description="List of code signature validation errors")
```
trevor-e, line 41: Same with this -> `code_signature_errors`, or can alternatively place in a nested model.

trevor-e (approved):

PR #52, opened 2025-06-19
Add a platform independent size analysis method

chromy:
This adds a new method:

do_size(input_file: BinaryIO) -> BaseAnalysisResult

- Which takes a file-like input containing an artifact
- Uses ArtifactFactory to create an Artifact
- Creates the correct Analyzer
- Returns the result of the analyzer

The point of this is to minimize the amount of boiler plate required
in getsentry/sentry and the launchpad service when adding a new kind of
artifact to be analyzed. Ideally the code path for
{Android,Apple,Unity,.so,.exe,...} would be almost identical until it
gets to `do_size`.

Also add a matching CLI command for this `launchpad size`. This is done
in a separate file (src/launchpad/size/cli.py) showing how we could
split up the subcommand definitions and have them live near the code
they use as opposed to all have to be in a large monolithic file.

While here also do a few other small cleanups:
- Move `Console()` into `console.py` so we make only one instance
  for the whole project (as suggested in the rich docs).
- Fix a few incorrect references to Android in artifact_factory.
- Add the ability for ArtifactFactory to take file-like objects as
  input.


src/launchpad/size/analysis.py
```diff
@@ -0,0 +1,28 @@
...
+from ..analyzers.apple import AppleAppAnalyzer
+from ..artifacts import AndroidArtifact, AppleArtifact, ArtifactFactory
+
+
+def do_size_analysis(input_file: BinaryIO, output_file: TextIO) -> None:
+    start_time = time.time()
+    artifact = ArtifactFactory.from_file(input_file)
+
+    # isinstance switch below is a bit sad. Ryan suggested a
+    # get_analyzer method on artifact which might be nicer.
```
rbro112, line 15: We can't do an `Artifact.get_analyzer` unfortunately due to dependency cycles (i.e. each artifact impl depends on `Artifact`, so `Artifact` cannot import implementations. I think for now this is our best bet, but we can rethink this longer term

src/launchpad/size/analysis.py
```diff
@@ -0,0 +1,28 @@
...
+    elif isinstance(artifact, AppleArtifact):
+        analyzer = AppleAppAnalyzer()
+    else:
+        raise ValueError(f"Unknown artifact kind {artifact}")
+    results = analyzer.analyze(cast(Any, artifact))
+
+    end_time = time.time()
+    duration = end_time - start_time
+    results = results.model_copy(update={"analysis_duration": duration})
+    json.dump(results.to_dict(), output_file, indent=2, ensure_ascii=False)
```
rbro112, line 28: I'd say we return `results` from this and the caller can then decide to write to JSON or dump to logs, etc. That way both CLI and future server usages could leverage this and decide what to do with the output

src/launchpad/size/cli.py
```diff
@@ -0,0 +1,60 @@
...
+@click.argument("input_path", type=click.Path(exists=True, path_type=Path), metavar="INPUT_PATH")
+@click.option(
+    "-o",
+    "--output",
+    type=click.File("w"),
+    help="Output path for the JSON analysis report (default: stdout).",
+)
+@click.option("--verbose", "-v", is_flag=True, help="Enable verbose logging output.")
+@click.option("--quiet", "-q", is_flag=True, help="Suppress all output except errors.")
+def size_command(
```
rbro112, line 22: I like this pattern of the subcommand living here 👍

src/launchpad/size/cli.py
```diff
@@ -0,0 +1,60 @@
...
+                    TextColumn("[progress.description]{task.description}"),
+                    console=console,
+                    disable=quiet,
+                )
+            )
+
+            input_file = stack.enter_context(open(input_path, "rb"))
+
+            task = progress.add_task("Analyzing...", total=None)
+            do_size_analysis(input_file, output)
```
rbro112, line 55: I think we lose the table printing with this pattern that I found useful. Personally I think the caller should determine what to do with the results rather than assuming the results should always be written to JSON

src/launchpad/size/cli.py
```diff
@@ -0,0 +1,60 @@
...
+import click
+from rich.progress import Progress, SpinnerColumn, TextColumn
+
+from ..utils import console, setup_logging
+from .analysis import do_size_analysis
+
+
+@click.command(name="size")
+@click.argument("input_path", type=click.Path(exists=True, path_type=Path), metavar="INPUT_PATH")
+@click.option(
```
rbro112, line 14: We lost `output_format` here, which I personally really liked and leveraged alot. See my comments [here](https://github.com/getsentry/launchpad/pull/52/files#r2157657809) and [here](https://github.com/getsentry/launchpad/pull/52/files#r2157661579) about suggestions to return the results and allow the caller to determine what to do with them.

src/launchpad/utils/__init__.py
```diff
@@ -1,6 +1,7 @@
 """Utility modules for app size analyzer."""
 
 from .android.bundletool import Bundletool, DeviceSpec
+from .console import console
```
rbro112, line 4: Not related to this change, but Trevor and I were chatting about the significance/usefulness of these __init__ exports. I know you have far more Python experience than us, if you know - what is the general pattern we should be following WRT exports in these files?

src/launchpad/cli.py
```diff
@@ -411,6 +409,9 @@ def _format_bytes(size: int) -> str:
     return f"{size_float:.1f} TB"
 
 
+cli.add_command(size_command)
```
rbro112, line 412: Seems we should cut out the existing CLI `android` && `ios` subcommands in favor of this new `size` command?

src/launchpad/size/analysis.py
chromy: Thanks! Removed this comment.

src/launchpad/size/analysis.py
chromy: Switched to returning the results directly.

src/launchpad/size/cli.py
chromy: Switched to returning the results directly.

src/launchpad/size/runner.py
```diff
@@ -0,0 +1,34 @@
...
+        analyzer = AndroidAnalyzer(**flags)
+    elif isinstance(artifact, AppleArtifact):
+        analyzer = AppleAppAnalyzer(**flags)
+    else:
+        raise ValueError(f"Unknown artifact kind {artifact}")
+    results = analyzer.analyze(cast(Any, artifact))
+
+    end_time = time.time()
+    duration = end_time - start_time
+    results = results.model_copy(update={"analysis_duration": duration})
```
trevor-e, line 29: I think this is already being computed inside each analyzer still, which should probably be removed in favor of this.

trevor-e (approved):
Thanks for cleaning this up!

src/launchpad/size/runner.py
chromy: Thanks! Sent a follow up for this one: https://github.com/getsentry/launchpad/pull/52#discussion_r2162675410

PR #53, opened 2025-06-19
Small Makefile fixes

chromy:
- Update `test` to invoke pytest on all tests. This:
  - saves paying the initaliztion cost of pytest twice
  - makes the output less visually noisy
- Remove `format` - it was an exact duplicate of `fix`.
- Rename `autofix` -> `fix`
- Rename `type-check` -> `check-types` (matching `check-format`)
- Remove `fix` from `check`. The idea being there should be:
  - one command which runs all the checks but *doesn't* modify any
    files. (`check` in this Makefile)
  - one command which autofixes all the checks (as far as possible).
    `fix` in this Makefile.
- Now we have some integration tests so remove the logic which
  displayed a message if there were none.

### DIFF
--- a/.cursor/rules/launchpad.mdc
+++ b/.cursor/rules/launchpad.mdc
@@ -61,7 +61,7 @@ This is a Python CLI tool for analyzing iOS and Android app bundle sizes, simila
 - `make help` - Show all available commands
 - `make dev-setup` - Set up development environment
 - `make test` - Run all tests (`test-unit`, `test-integration` available separately)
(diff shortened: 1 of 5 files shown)


chromy:
On the renames in particular I was trying to make them a bit more consistent but:
a) did not love what I ended up with
and b) I don't have that strong opinions so fine to put them back/change to something else.

NicoHinderling (approved):

runningcode (approved):

PR #65, opened 2025-06-24
Add cwebp for future size analysis on CI

runningcode:
Copying the pattern from bundletool

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,6 +76,16 @@ jobs:
           echo "$(pwd)" >> $GITHUB_PATH
           ./bundletool version
 
+      - name: Setup cwebp
+        run: |
+          curl -L https://storage.googleapis.com/downloads.webmproject.org/releases/webp/libwebp-1.5.0-linux-x86-64.tar.gz -o webp.tar.gz
+          tar -xzf webp.tar.gz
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -80,6 +80,13 @@ jobs:
           echo "$(pwd)" >> $GITHUB_PATH
           ./bundletool version
 
+      - name: Setup cwebp
+        run: |
+          curl -L https://storage.googleapis.com/downloads.webmproject.org/releases/webp/libwebp-1.5.0-linux-x86-64.tar.gz -o webp.tar.gz
+          tar -xzf webp.tar.gz
+          echo "$(pwd)/libwebp-1.5.0-linux-x86-64/bin" >> $GITHUB_PATH
+          ./libwebp-1.5.0-linux-x86-64/bin/cwebp -version
```
runningcode, line 88: weirdly the $GITHUB_PATH seems to only apply to the next step and not to this one so we have to call it directly.
seems like the same pattern was used for bundletool

runningcode: Ah i was trying to answer the question you asked before you asked it but I didn’t explain it well. Maybe my comment below helps with the link to the example? https://github.com/getsentry/launchpad/pull/65/files#r2166018284

PR #66, opened 2025-06-24
update devservices to have the correct kafka topic name

NicoHinderling:
forgot to update this too as part of https://github.com/getsentry/launchpad/pull/61


chromy (approved):

Makefile
```diff
@@ -10,11 +10,14 @@ VENV_DIR := .venv
 UV := uv
 PYTHON_VENV := $(VENV_DIR)/bin/python
 
-# Kafka configuration for local development
+# local development environment variables
```
hubertdeng123, line 13: I believe these env variables may be better placed into your .envrc file, that way they are always loaded by direnv as long as you are in the launchpad directory

PR #78, opened 2025-06-25
Add Webp Optimization analysis

runningcode:
This checks to see if the images could be smaller if they were optimized
with webp.
This uses the same logic as we previously had in
`WebPAssetInsightDetector.kt`

Constraints:
* only scan `png`, `jpeg`, `jpg` or `bmp`
* skip 9 patch (`.9.png`)
* only scan `res` or `asset` directories

There's a lot of code in here so I'm looking to make sure the direction
is correct.

Questions:
* Do we have a pattern that I should use for creating the temporary
  files to scan the webp?
* Do we need the unit tests? Or are the integration tests sufficient?
* I added an Artifact field to `InsightsInput` is there a better way to
  do this?

### DIFF
--- a/src/launchpad/analyzers/android.py
+++ b/src/launchpad/analyzers/android.py
@@ -9,6 +9,7 @@ from ..artifacts.android.zipped_apk import ZippedAPK
 from ..artifacts.artifact import AndroidArtifact
 from ..insights.common import DuplicateFilesInsight
 from ..insights.insight import InsightsInput
+from ..insights.webp_asset import WebpAssetInsight
 from ..models.android import (
     AndroidAnalysisResults,
     AndroidAppInfo,
@@ -82,9 +83,11 @@ class AndroidAnalyzer:
                 file_analysis=file_analysis,
                 treemap=treemap,
                 binary_analysis=[],
+                artifact=artifact,
             )
             insights = AndroidInsightResults(
                 duplicate_files=DuplicateFilesInsight().generate(insights_input),
+                webp_assets=WebpAssetInsight().generate(insights_input),
             )
 
         return AndroidAnalysisResults(
--- a/src/launchpad/insights/webp_asset.py
+++ b/src/launchpad/insights/webp_asset.py
@@ -1,0 +1,246 @@
+"""WebP asset optimization insights for Android apps."""
+
+from __future__ import annotations
+
+import tempfile
+
+from pathlib import Path
+from typing import Any, Dict, List, Optional
+
+from ..models.common import FileInfo
+from ..models.insights import WebpAssetInsightResult
+from ..utils.image.cwebp import Cwebp
+from ..utils.logging import get_logger
+from .common import Insight, InsightsInput
+
+logger = get_logger(__name__)
+
+SUPPORTED_IMAGE_FORMATS = {".png", ".jpg", ".jpeg", ".tiff", ".tif", ".bmp"}
+
+MIN_IMAGE_SIZE = 4 * 1024
+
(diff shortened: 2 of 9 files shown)


runningcode:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/getsentry/launchpad/78?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#78** <a href="https://app.graphite.dev/github/pr/getsentry/launchpad/78?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/getsentry/launchpad/78?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#65** <a href="https://app.graphite.dev/github/pr/getsentry/launchpad/65?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#74](https://github.com/getsentry/launchpad/pull/74) <a href="https://app.graphite.dev/github/pr/getsentry/launchpad/74?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

src/launchpad/analyzers/android.py
```diff
@@ -82,9 +83,11 @@ def analyze(self, artifact: AndroidArtifact) -> AndroidAnalysisResults:
                 file_analysis=file_analysis,
                 treemap=treemap,
                 binary_analysis=[],
+                artifact=artifact,
```
runningcode, line 86: Let me know if there’s a better way to do this. We need to open the files from the artifact in order to get access to the images.

src/launchpad/insights/webp_asset.py
```diff
@@ -0,0 +1,246 @@
...
+from ..utils.logging import get_logger
+from .common import Insight, InsightsInput
+
+logger = get_logger(__name__)
+
+SUPPORTED_IMAGE_FORMATS = {".png", ".jpg", ".jpeg", ".tiff", ".tif", ".bmp"}
+
+MIN_IMAGE_SIZE = 4 * 1024
+
+MIN_SAVINGS_THRESHOLD = 500
```
runningcode, line 22: copied from `WebPAssetInsightDetector.kt`

runningcode (commented):
Added some questions to guide the review. i have a lot of questions since I’m quite new at python

src/launchpad/models/insights.py
```diff
@@ -22,3 +22,14 @@ class DuplicateFilesInsightResult(BaseInsightResult):
     def duplicate_count(self) -> int:
         """Number of duplicate files (excluding the original)."""
         return len(self.files) - 1
+
+
+class WebpAssetInsightResult(BaseInsightResult):
+    """Results from WebP asset optimization analysis."""
```
rbro112, line 28: Can you remove these comments, they're not helpful IMO

src/launchpad/analyzers/android.py
rbro112: I don't think this is the pattern we want to follow. Both treemap/insights shouldn't know about the artifact. The analyzer is responsible for taking the artifact and doing the "analysis" on it, then generating the treemap/insights from that analysis. I think a separate function in `android.py` here to find all images and check if they can be optimized, then return the list of optimizable images that the InsightsInput would take would be better.

cc @trevor-e as we've been discussing this a bit offline

PR #90, opened 2025-06-27
Outline of function for image optimization insight

runningcode:


tests/unit/test_image_optimization_insight.py
```diff
@@ -0,0 +1,131 @@
...
+
+from launchpad.analyzers.android import AndroidAnalyzer
+from launchpad.models.insights import ImageOptimizationInsightResult
+
+
+class TestImageOptimizationInsight:
+    def setup_method(self):
+        self.analyzer = AndroidAnalyzer()
+
+    @patch("launchpad.utils.android.cwebp.Cwebp.convert_to_webp")
```
runningcode, line 18: not sure if these tests make so much sense to keep since everything is mocked.
but since it is hard to do a full integration test unless we add some more large files to the repo, maybe this is best. open to thoughts!

PR #95, opened 2025-06-27
Add asset catalog parsing

noahsmartin:


src/launchpad/analyzers/apple.py
```diff
@@ -297,18 +303,60 @@ def _analyze_files(self, xcarchive: ZippedXCArchive) -> FileAnalysis:
...
                 hash_md5=file_hash,
                 treemap_type=FILE_TYPE_TO_TREEMAP_TYPE.get(file_type, TreemapType.FILES),
+                children=children,
             )
 
             files.append(file_info)
 
         return FileAnalysis(files=files)
 
+    def _analyze_asset_catalog(self, xcarchive: ZippedXCArchive, relative_path: Path) -> List[TreemapElement]:
```
trevor-e, line 336: Check out the pattern that Ryan has started for us:
```
    def _create_file_element(self, file_info: FileInfo, display_name: str) -> TreemapElement:
        default_element_builder = DefaultFileElementBuilder(
            download_compression_ratio=self.download_compression_ratio,
            filesystem_block_size=self.filesystem_block_size,
        )

        element_builder: TreemapElementBuilder = default_element_builder
        match file_info.file_type:
            case "macho":
                element_builder = MachOElementBuilder(
                    binary_analysis_map=self.binary_analysis_map,
                    download_compression_ratio=self.download_compression_ratio,
                    filesystem_block_size=self.filesystem_block_size,
                )
            case "dex":
                element_builder = DexElementBuilder(
                    class_definitions=self.class_definitions,
                    download_compression_ratio=self.download_compression_ratio,
                    filesystem_block_size=self.filesystem_block_size,
                )
```
I think this is a pretty scalable way to add more and more file types that can have nested nodes.

src/launchpad/analyzers/apple.py
```diff
@@ -297,18 +303,60 @@ def _analyze_files(self, xcarchive: ZippedXCArchive) -> FileAnalysis:
...
+        """Analyze an asset catalog file."""
+        catalog_details = xcarchive.get_asset_catalog_details(relative_path)
+        return [
+            TreemapElement(
+                name=element.name,
+                install_size=element.size,
+                # TODO: This field should be nullable, it doesn’t make sense
```
trevor-e, line 343: Could we take the ratio of the asset catalog size to the sum size of all the extracted children and then use that as the download size ratio per image?

trevor-e (approved):
Stamped to unblock but would check out the pattern I mentioned

src/launchpad/analyzers/apple.py
noahsmartin: I think we should stick to only what we can back up with real data, that wouldn't be the download size, it would just be some number. The only thing I could think of would be if we could somehow remove each image one at a time and compute the download size without that image present. Even that wouldn't be very helpful to users though because it's sum could be more than the real download size

src/launchpad/analyzers/apple.py
noahsmartin: I think this is essentially the same thing but instead of having the node know its own builder it has to be passed through this `TreemapBuilder` type which isn't quite as flexible. For example if we created a new AssetCatalogBuilder and copied the current logic into that, it wouldn't have access to the xcarchive in the `TreemapBuilder` class. Maybe we could have the `ElementBuilder` be a property on the `FileInfo`, but that's pretty similar to what this PR does by just attaching the tree directly, no need to really have a separate `ElementBuilder` type.

So TLDR: With this way I don't see a good way to pass the necessary info to the "element builder" it would all have to get piped through the `TreemapBuilder` initializer

PR #97, opened 2025-06-30
Add scripts/deps to download dependencies

chromy:
Launchpad requires a bunch of tools at runtime. The current intended
list is: `ffmpeg`, `assetutil`, `imagemin`, `afconvert`, `codesign`, 
`lipo`, `bitcode_strip`, `strip`, `dyld_info`, `apksigner`, `bundletool`,
`aapt2`, `cwebp` - but more may be discovered.

We need these tools in three scenarios:
- local development 
- CI
- production

Ideally you want to use a) a pinned version and b) the same pinned
version across all three situations. https://github.com/getsentry/launchpad/pull/86
changed the tests to use Docker which solves keeping CI and
production in sync - but still means for local development you
either need to always build the docker image (which is slow) or
manually install deps as you discover them missing.

Also while testing the Docker image in CI is good (in that it is
more representative of production) there are some cases where
it becomes a hassle to e.g. compute coverage https://github.com/getsentry/launchpad/pull/97
which was the motivation for this change.

This adds a script which downloads all the needed dependancies
into:
- `.devenv/*`

We divide the dependancies into  
- `.devenv/all/` - for platform independent tools (e.g. `.jar`s)
- `.devenv/x86_64-linux/`
- `.devenv/aarch64-darwin/`

etc - for native binaries.

We take the actual tools from the downloaded files and copy them to:
- `.devenv/all/bin/`
- `.devenv/x86_64-linux/bin/`
- `.devenv/aarch64-darwin/bin/`

Adding a tool is simple:

```
Dep(
      "mynativetool.tar.gz",
      "https://github.com/example/example/archive/refs/tags/mynativetool.tar.gz",
      "f4bf49f85991f50e86a5404d16f15b72a053bb66768ed5cc0f6d042277cc2bb8",
      architecture="x86_64",
      system="linux",
      binaries=[
          Bin(source="mynativetool/foo", target="foo"),
      ],
),
```
for a platform dependent tool. Or just:
```
Dep(
    "bundletool.jar",
    "https://github.com/google/bundletool/releases/download/1.18.1/bundletool-all-1.18.1.jar",
    "675786493983787ffa11550bdb7c0715679a44e1643f3ff980a529e9c822595c",
),
```
for an platform independent tool.

- The hash ensures everyone is always using the exact same
version and makes it easy to detect when we need to download
a new version.
- It's fully hermetic so you're never in a situation where you are
breaking your global dev environment because `launchpad`
needs an older/newer version of `bundletool` than something
else.

This allows us to:
- make Docker image creation more hermetic and simpler
- gets closer to 'single button' dev env setup.

This script is used from:
- `make` - as part of `make install-deps` making it useful again.
- `devenv` - as part of `devenv sync`

In `.envrc` we add the relevant paths to `PATH`.
In the `Dockerfile` we copy the relevant files into the image.

Rather than having a separate script it would have been nice to add
this to `devenv` which broadly seems to be solving a similar issue
but just does not want to work in CI.

### DIFF
--- a/scripts/deps
+++ b/scripts/deps
@@ -1,0 +1,313 @@
+#!/usr/bin/env python
+
+"""
+Handles downloading and ensuring the runtime dependencies of launchpad
+are up-to-date. We require various dependencies
+(cwebp, bundletool, etc). Some of these are architecture/OS dependant
+and some are not. There are three situations to care about:
+local development, CI, and production. We ideally want use the same
+pinned version of the dependencies in three situations and specify
+those versions in a single place.
+
+This script can be invoked from: devenv, make, and, manually.
+"""
+
+from dataclasses import dataclass
+from typing import Optional, List
+import argparse
+import os
+import platform
+import sys
+import hashlib
+import subprocess
+import shutil
+import stat
+import textwrap
+
+# Don't add non-standard library dependencies since this can be ran
+# prior to the venv getting set up.
+
+ROOT_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+DEPS_DIR = os.path.join(ROOT_DIR, ".devenv")
+KNOWN_ARCHITECTURES = {"x86_64", "aarch64"}
+KNOWN_SYSTEMS = {"darwin", "linux"}
+
+
+@dataclass
+class Bin:
+    text: Optional[str] = None
+    source: Optional[str] = None
+    target: str = ""
+
+    def __post_init__(self):
+        assert bool(self.text) != bool(self.source), f"{self} must define text xor source"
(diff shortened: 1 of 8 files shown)


scripts/deps
```diff
@@ -0,0 +1,311 @@
...
+
+
+    bad_deps = get_bad_deps()
+
+    if not bad_deps:
+        return 0
+
+    if args.check:
+        argz = " ".join([a for a in sys.argv[1:] if a != "--check"])
+        print(f"\033[91mBuild deps are stale. Please run: scripts/deps {argz}\033[0m")
```
noahsmartin, line 267: Is this not actually doing any checking? Should we delete the check flag?

PR #105, opened 2025-06-30
Use file offsets when calculating symbol size

trevor-e:
I haven't noticed many differences so far when inspecting the top symbol sizes, but this is more technically correct than using the virtual address sizes.


src/launchpad/parsers/apple/macho_symbol_sizes.py
```diff
@@ -38,6 +38,7 @@ def get_symbol_sizes(self) -> list[SymbolSize]:
             )
 
         logger.info(f"Found {len(symbol_sizes)} symbol sizes")
+        symbol_sizes.sort(key=lambda x: x.size, reverse=True)
```
trevor-e, line 41: Added as a convenience

PR #109, opened 2025-07-01
Replace kafka consumer with more arroyo logic

NicoHinderling:
Arroyo implements a lot of what I manually implemented + handles corner cases that I didn't have

### DIFF
--- a/src/launchpad/service.py
+++ b/src/launchpad/service.py
@@ -7,7 +7,7 @@ import os
 import signal
 import time
 
-from typing import Any, Dict
+from typing import Any, Dict, cast
 
 from arroyo.backends.kafka import KafkaPayload
 from arroyo.processing.processor import StreamProcessor
@@ -191,11 +191,14 @@ class LaunchpadService:
 
     async def health_check(self) -> HealthCheckResponse:
         """Get overall service health status."""
-        health_status: HealthCheckResponse = {
-            "service": "launchpad",
-            "status": "ok",
-            "components": {},
-        }
+        health_status = cast(
+            HealthCheckResponse,
+            {
+                "service": "launchpad",
+                "status": "ok",
+                "components": {},
+            },
+        )
 
         # Check Kafka health via healthcheck file
         kafka_health: Dict[str, Any] = {"status": "unknown"}


src/launchpad/kafka.py
```diff
@@ -29,234 +29,92 @@
...
-        logger.info("Kafka consumer run loop completed")
+                logger.error(f"Failed to decode message: {e}")
+                raise  # Re-raise the exception to prevent processing invalid messages
+
+        return RunTaskInThreads(
+            processing_function=process_message,
+            concurrency=self.concurrency,
+            max_pending_futures=self.max_pending_futures,
+            next_step=next_step,
+        )
```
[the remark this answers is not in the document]
john-z-yang: The message (the metadata bit) actually gets passed to the next step, so that commit step will know what offset to commit to the kafka broker.

The kafka message itself gets enqueued into a queue along side the future that will yield the result of the function call [here](https://github.com/getsentry/arroyo/blob/main/arroyo/processing/strategies/run_task_in_threads.py#L84). And gets passed to the next step [here](https://github.com/getsentry/arroyo/blob/main/arroyo/processing/strategies/run_task_in_threads.py#L121)

john-z-yang (approved):
I don't have a lot of context on `service.py` but the consumer part looks good to me.

Assuming you're using sentry's `devservice`. If you want to test locally, you can run
```bash
docker exec -it kafka-kafka-1 bash
```
to ssh into the kafka broker container, and run
```bash
kafka-consumer-groups --bootstrap-server localhost:9092 --describe --group your_consumer_group_id
```
To have it print out the offsets, so you can make sure that offsets are committed and you're maintaining at-least-once guarantee.

src/launchpad/kafka.py
```diff
@@ -29,234 +29,92 @@
...
+
+
+def create_kafka_consumer(
```
trevor-e, line 77: nit: can `create_kafka_consumer` be first in the file? seems like `LaunchpadStrategyFactory` isn't used externally anywhere.

src/launchpad/kafka.py
```diff
@@ -29,234 +29,92 @@
...
-        """Check the health of the Kafka connection."""
-        return {
-            "status": "running" if not self._shutdown_requested else "stopped",
-            "topics": self.topics,
-            "group_id": self.group_id,
-            "bootstrap_servers": self.bootstrap_servers,
-            "processor_active": self._processor is not None,
-        }
+    # Create and return stream processor
+    topics = [Topic(topic) for topic in config["topics"]]
```
trevor-e, line 107: New to this area but what's the purpose of throwing away all the topics other than `topics[0]`?

src/launchpad/server.py
```diff
@@ -6,7 +6,7 @@
 import logging
 import os
 
-from typing import Any, Dict
+from typing import Any, Callable, Dict, Optional
```
trevor-e, line 9: nit: use `| None` instead of Optional

src/launchpad/server.py
```diff
@@ -107,15 +109,35 @@ async def create_app(self) -> Application:
...
-            }
-        )
+        """Health check endpoint that checks all service components."""
+        try:
+            # Get health status from the service if callback is provided
+            if self.health_check_callback:
+                health_data = await self.health_check_callback()
+
+                # Determine HTTP status code based on health status
+                if health_data.get("status") == "ok":
```
trevor-e, line 119: Does this need more resilient handling or will it always handle JSON data? I believe this is typed as `Any` and then we are assuming it returns JSON.

trevor-e (approved):

src/launchpad/kafka.py
NicoHinderling: I think this is related to arroyo's design philosphy, specifically for the StreamProcessor. It's meant to only process messages from one topic with a specific processing strategy.

I think
```
    topics = [Topic(topic) for topic in config["topics"]]
```
should only return an array of size 1 just for our topic, since `config["topics"]` will equal to `preprod-artifact-events`, our topic name

PR #119, opened 2025-07-03
Stop CodeCov annotations

chromy:
See:
https://docs.codecov.com/docs/github-checks#disabling-github-checks-patch-annotations-via-yaml

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,3 @@
+github_checks:
+    annotations: false
+


runningcode (approved):

runningcode:
Just so I understand, this disables them but then you can re-enable them via the menu?

chromy:
> Just so I understand, this disables them but then you can re-enable them via the menu?

No I think this prevents them being added in the first place. Github generally allows you to toggle annotations on and off via `a` which the docs suggest as an alternative to this - but I think few people know that is possible (I didn't before reading the docs for example).

This doesn't change the github comment at all and the coverage on the diff should be visible in the linked codecov report.

PR #127, opened 2025-07-07
Add Hermes bytecode parser for iOS/Android

trevor-e:
Screenshot 2025-07-07 at 5 37 32 PM


src/launchpad/parsers/buffer_wrapper.py
```diff
@@ -29,13 +31,14 @@ def __exit__(
...
-        logger.debug("=== End %s ===", self.name)
+        if self.enabled:
+            logger.debug("=== End %s ===", self.name)
 
 
 class BufferWrapper:
     """Wrapper for binary buffer parsing with cursor tracking and debugging."""
 
-    def __init__(self, buffer: bytes) -> None:
+    def __init__(self, buffer: bytes, debug: bool = False) -> None:
```
trevor-e, line 41: Had to add this in because otherwise parsing takes forever locally. If you need it for debugging then can manually set.

rbro112: Fine with me, agree that the debug logging was out of hand without this

rbro112 (approved):

PR #131, opened 2025-07-08
Simplify webp analysis by adding fully qualified path to FileInfo

runningcode:
This should make the code simpler and improve performance because having to iterate through all the files for every image found would be quite slow.

### DIFF
--- a/src/launchpad/size/analyzers/android.py
+++ b/src/launchpad/size/analyzers/android.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 from datetime import datetime, timezone
-from pathlib import Path
 
 from launchpad.artifacts.android.aab import AAB
 from launchpad.artifacts.android.apk import APK
@@ -97,7 +96,6 @@ class AndroidAnalyzer:
                 file_analysis=file_analysis,
                 treemap=treemap,
                 binary_analysis=[],
-                image_map=self._get_images(apks, file_analysis),
             )
             insights = AndroidInsightResults(
                 duplicate_files=DuplicateFilesInsight().generate(insights_input),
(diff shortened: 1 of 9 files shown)


runningcode:
* **#131** <a href="https://app.graphite.dev/github/pr/getsentry/launchpad/131?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/getsentry/launchpad/131?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/launchpad/size/analyzers/android.py
```diff
@@ -243,14 +245,3 @@ def _get_hermes_reports(self, apks: list[APK]) -> dict[str, HermesReport]:
                     logger.warning(f"Duplicate Hermes report key found: {relative_path}, overwriting")
                 all_reports[relative_path] = report
         return all_reports
-
-    def _get_images(self, apks: list[APK], file_analysis: FileAnalysis) -> dict[Path, FileInfo]:
-        image_map = {}
-        for apk in apks:
-            for image_file in apk.get_images():
-                image_map[image_file] = next(
-                    (file_info for file_info in file_analysis.files if str(image_file).endswith(file_info.path)), None
```
runningcode, line 252: This change should also improve performance as having to iterate through all the files would have been very slow.

src/launchpad/size/models/common.py
```diff
@@ -60,11 +60,12 @@ class FileInfo(BaseModel):
     model_config = ConfigDict(frozen=True)
 
     path: str = Field(..., description="Relative path in the bundle")
+    full_path: str = Field(..., description="Fully qualified path to the file")
```
trevor-e, line 63: I was looking into this yesterday for a similar reason to yours. One thought: maybe this should use `exclude=True` so it doesn't get serialized later on? I think this can have a pretty big size effect on the resulting JSON file, and we only need the relative path for external use.

trevor-e (approved):

PR #142, opened 2025-07-08
Wire up end to end call path (kafka -> monolith endpoints)

NicoHinderling:
A kafka message will now trigger the following flow:
1. try to download the corresponding artifact
2. preprocess it
3. send the preprocess data to the monolith
4. full size analysis
5. send the full size analysis file to the monolith

### DIFF
--- a/src/launchpad/sentry_client.py
+++ b/src/launchpad/sentry_client.py
@@ -11,7 +11,7 @@ import re
 import secrets
 
 from pathlib import Path
-from typing import Any, Dict, NamedTuple
+from typing import Any, Dict, NamedTuple, cast
 
 import requests
 
@@ -75,7 +75,7 @@ class SentryClient:
 
         self.session = create_retry_session()
 
-    def download_artifact_to_file(self, org: str, project: str, artifact_id: str, out) -> int | ErrorResult:
+    def download_artifact_to_file(self, org: str, project: str, artifact_id: str, out) -> int:
         """Download preprod artifact directly to a file-like object.
 
         Args:
@@ -85,39 +85,33 @@ class SentryClient:
             out: File-like object to write to (must support write() method)
 
         Returns:
-            Number of bytes written on success, or ErrorResult on failure
+            Number of bytes written on success
+
+        Raises:
+            RuntimeError: With categorized error message on failure
         """
         endpoint = f"/api/0/internal/{org}/{project}/files/preprodartifacts/{artifact_id}/"
         url = self._build_url(endpoint)
 
         logger.debug(f"GET {url}")
+
         response = self.session.get(url, headers=self._get_auth_headers(), timeout=120, stream=True)
 
         if response.status_code != 200:
-            return self._handle_error_response(response, "Download")
+            error_result = self._handle_error_response(response, "Download artifact")
+            error_category, error_description = categorize_http_error(error_result)
+            raise RuntimeError(f"Failed to download artifact ({error_category}): {error_description}")
 
         # Stream directly to the file-like object
         file_size = 0
-        try:
-            for chunk in response.iter_content(chunk_size=8192):
-                if chunk:
-                    out.write(chunk)
-                    file_size += len(chunk)
-                    if file_size > 5 * 1024 * 1024 * 1024:  # 5GB limit
-                        logger.error("Download exceeds 5GB limit")
-                        return ErrorResult(
-                            error="File size exceeds 5GB limit",
-                            status_code=413,  # Payload Too Large
-                        )
+        for chunk in response.iter_content(chunk_size=8192):
+            if chunk:
+                out.write(chunk)
+                file_size += len(chunk)
+                if file_size > 5 * 1024 * 1024 * 1024:  # 5GB limit
+                    raise RuntimeError("Failed to download artifact (client_error): File size exceeds 5GB limit")
 
-            return file_size
-
-        except Exception as e:
-            logger.error(f"Failed to write to file-like object: {e}")
-            return ErrorResult(
-                error=f"Failed to write to file-like object: {e}",
-                status_code=500,
-            )
+        return file_size
 
     def update_artifact(
         self, org: str, project: str, artifact_id: str, data: Dict[str, Any]
@@ -228,9 +222,11 @@ class SentryClient:
     def _handle_error_response(self, response: requests.Response, operation: str) -> ErrorResult:
         """Handle non-200 response with consistent error format."""
         logger.warning(f"{operation} failed: {response.status_code}")
+        # Cast to int to help type checker understand status_code is int
+        status_code = cast(int, response.status_code)
         return ErrorResult(
-            error=f"HTTP {response.status_code}",
-            status_code=response.status_code,
+            error=f"HTTP {status_code}",
+            status_code=status_code,
         )
 
     def _make_json_request(
--- a/src/launchpad/service.py
+++ b/src/launchpad/service.py
@@ -170,11 +170,12 @@ class LaunchpadService:
                 data=update_data,
             )
 
-            if "error" in update_result:
-                error_msg = f"Failed to send preprocessed info: {update_result['error']}"
+            # Check if update_result is an ErrorResult
+            if isinstance(update_result, ErrorResult):
+                error_category, error_description = categorize_http_error(update_result)
+                error_msg = f"Failed to send preprocessed info: {error_description}"
                 logger.error(error_msg)
-                # Extract just the error details for the database, the enum prefix will be added automatically
-                detailed_error = update_result["error"]
+                # Use the categorized error description for the database
                 self._update_artifact_error(
                     sentry_client,
                     artifact_id,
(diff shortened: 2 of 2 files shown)


NicoHinderling:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/getsentry/launchpad/142?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#142** <a href="https://app.graphite.dev/github/pr/getsentry/launchpad/142?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/getsentry/launchpad/142?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#141** <a href="https://app.graphite.dev/github/pr/getsentry/launchpad/141?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#140** <a href="https://app.graphite.dev/github/pr/getsentry/launchpad/140?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/launchpad/size/analyzers/android.py
```diff
@@ -48,17 +48,35 @@ def __init__(
...
         )
 
+        return self.app_info
+
+    def analyze(self, artifact: AndroidArtifact) -> AndroidAnalysisResults:
+        # Use preprocessed app info if available, otherwise extract it
+        if not self.app_info:
+            self.app_info = self.preprocess(artifact)
+
+        app_info = self.app_info
```
NicoHinderling, line 78: Now matches analyzers/apple.py

src/launchpad/sentry_client.py
```diff
@@ -27,36 +49,34 @@ def __init__(self, base_url: str) -> None:
...
+        if response.status_code != 200:
+            return self._handle_error_response(response, "Download")
+
+        # Read content with size limit
+        content = b""
+        for chunk in response.iter_content(chunk_size=8192):
+            if chunk:
+                content += chunk
+            if len(content) > 5 * 1024 * 1024 * 1024:  # 5GB limit
+                logger.warning("Download truncated at 5GB")
```
noahsmartin, line 71: Should we return an error here instead?

src/launchpad/service.py
```diff
@@ -78,58 +127,396 @@ def handle_kafka_message(self, payload: PreprodArtifactEvents) -> None:
...
 
-        # Start Kafka processor in a background thread
-        loop = asyncio.get_event_loop()
-        self._kafka_task = loop.run_in_executor(None, self.kafka_processor.run)
+            logger.info(f"Successfully sent preprocessed info for artifact {artifact_id}")
 
-        # Start HTTP server as a background task
-        server_task = asyncio.create_task(self.server.start())
+            analyzer = self._create_analyzer(app_info)
+            logger.info(f"Running full analysis on {temp_file}...")
```
noahsmartin, line 197: Eventually we'll want some field in the kafka message to indicate if size analysis is enabled

src/launchpad/service.py
```diff
@@ -78,58 +127,396 @@ def handle_kafka_message(self, payload: PreprodArtifactEvents) -> None:
...
             raise RuntimeError("Service not properly initialized. Call setup() first.")
 
-        logger.info("Starting Launchpad service...")
+        sentry_client = SentryClient(base_url=self._service_config["sentry_base_url"])
+        temp_file = None
 
-        # Set up signal handlers for graceful shutdown
-        self._setup_signal_handlers()
+        try:
+            file_content, _ = self._download_artifact(sentry_client, artifact_id, project_id, organization_id)
```
noahsmartin, line 158: I wonder if it's a problem that we are keeping this large file in memory and then writing it, and reading it all again (in ArtifactFactory). Ideally the download would write to the file without storing the entire thing in memory. Maybe we should at least make sure this `file_content` reference isn't held when we read it the second time?

src/launchpad/service.py
```diff
@@ -78,58 +127,396 @@ def handle_kafka_message(self, payload: PreprodArtifactEvents) -> None:
...
+                    artifact_id,
+                    project_id,
+                    organization_id,
+                    ERROR_CODE_ARTIFACT_PROCESSING_ERROR,
+                    ProcessingErrorMessage.DOWNLOAD_FAILED,
+                    detailed_error,
+                )
+            raise
+
+    def _save_to_temp_file(self, file_content: bytes, artifact_id: str) -> str:
```
chromy, line 388: The handling of tempfile is a little odd. Ideally we would use the built in cleanup behaviour of tempfile rather than manually cleaning up ourselves. See comment at use.

src/launchpad/sentry_client.py
```diff
@@ -27,36 +49,34 @@ def __init__(self, base_url: str) -> None:
...
+                break
+
+        return {
+            "success": True,
+            "file_content": content,
+            "file_size_bytes": len(content),
+            "headers": dict(response.headers),
+        }
```
chromy, line 79: The DX of dicts (both types and access) are quite bad for Python vs. TypeScript so what it very idiomatic in TypeScript (making the return type an anonymous object with some fields) is not the most common in Python.

If there are only ~2 fields it's more common to use a tuple:
```
def foo():
  ...
  success = True
  content = "bar"
  return success,  content
```

However if there are many fields the fastest thing is to use a `NamedTuple`:
```
class DownloadResult(NamedTuple):
    success: bool
    file_content: string
    file_size_bytes: int
```
Which is a magic way to define an object with some fields (also handles a bunch of the operators). See:

https://typing.python.org/en/latest/spec/namedtuples.html#named-tuples

src/launchpad/sentry_client.py
```diff
@@ -27,36 +49,34 @@ def __init__(self, base_url: str) -> None:
...
+            return self._handle_error_response(response, "Download")
+
+        # Read content with size limit
+        content = b""
+        for chunk in response.iter_content(chunk_size=8192):
+            if chunk:
+                content += chunk
+            if len(content) > 5 * 1024 * 1024 * 1024:  # 5GB limit
+                logger.warning("Download truncated at 5GB")
+                break
```
chromy, line 72: I think we should avoid reading the whole file into a string if possible since they can be quite large.
Python has a nice abstraction for file-like objects which is what I would use here.

We could either take the file-like object as an out parameter:

```
def download_artifact(..., out):
    for chunk in response.iter_content(chunk_size=8192):
        out.write(chunk)
```

Or you can have `download_artifact` make a tempfile/in memory file:

```
def download_artifact(...):
    f = tempfile.TemporaryFile()
    for chunk in response.iter_content(chunk_size=8192):
        f.write(chunk)
    return f
```

```
def download_artifact(...):
    f = io.StringIO()
    for chunk in response.iter_content(chunk_size=8192):
        f.write(chunk)
    return f
```

src/launchpad/service.py
```diff
@@ -78,58 +127,396 @@ def handle_kafka_message(self, payload: PreprodArtifactEvents) -> None:
...
+                )
+
+            file_content = download_result["file_content"]
+            file_size = download_result["file_size_bytes"]
+
+            logger.info(f"Downloaded artifact {artifact_id}: {file_size} bytes ({file_size / 1024 / 1024:.2f} MB)")
+            return file_content, file_size
+
+        except Exception as e:
+            if not isinstance(e, RuntimeError):
```
chromy, line 375: If you want to catch only some kinds of errors you can do:

```python
try:
   # ...
except RuntimeError as e:
   # handle RuntimeError here
```

src/launchpad/service.py
```diff
@@ -78,58 +127,396 @@ def handle_kafka_message(self, payload: PreprodArtifactEvents) -> None:
...
+        project_id: str,
+        organization_id: str,
+    ) -> tuple[bytes, int]:
+        """Download artifact from Sentry and validate response."""
+        logger.info(f"Downloading artifact {artifact_id}...")
 
         try:
-            # Wait for shutdown signal
-            await self._shutdown_event.wait()
+            download_result = sentry_client.download_artifact(
```
chromy, line 345: Here it might be better to do:

```python
with tempfile.TemporyFile() as tf:
  sentry_client.download_artifact(org=organization_id, project=project_id, artifact_id=artifact_id, out=tf)
  artifact = ArtifactFactory.fromFile(tf)
  ...
  
```

src/launchpad/service.py
```diff
@@ -78,58 +127,396 @@ def handle_kafka_message(self, payload: PreprodArtifactEvents) -> None:
...
+                self._handle_download_error(
+                    sentry_client,
+                    artifact_id,
+                    project_id,
+                    organization_id,
+                    f"Failed to download artifact ({error_category}): {error_description}",
+                )
+
+            if not download_result.get("success"):
+                self._handle_download_error(
```
chromy, line 360: The control flow here seems odd, `_handle_download_error` ends up rasing iiuc. Can `download_artifact` just raise an error directly?

src/launchpad/service.py
NicoHinderling: 👍 created a linear ticket for this https://linear.app/getsentry/issue/EME-84/add-a-field-in-the-kafka-message-to-indicate-what-type-of-analyses-to

PR #149, opened 2025-07-09
Use shutil.which in bundletool.py

chromy:


runningcode:
Oh, I already did this in https://github.com/getsentry/launchpad/pull/147

PR #151, opened 2025-07-09
Refactor dex parser to be more modular and implement class parsing

rbro112:
Modularize dex parsing to be per class/method/field, starting with classes. This matches the pattern set by the https://github.com/JesusFreke/smali library which I found much better to follow and easier to test/iterate on.


src/launchpad/parsers/buffer_wrapper.py
```diff
@@ -225,7 +225,7 @@ def read_sized_int(self, size: int) -> int:
             ValueError: If size is invalid
         """
         with self._debug_group(f"read_sized_int ({size} bytes)"):
-            if not 1 <= size <= 4:
+            if not 1 <= size <= 8:
```
rbro112, line 228: Intentional, longs don't exist in python: https://peps.python.org/pep-0237/

trevor-e (approved):

PR #158, opened 2025-07-10
Add Strip Symbols analysis for xcarchives.

runningcode:
This is based on the logic in `BundleAnalyzer.swift`.

I created a test fixture from the hacker news app by deleting the strip
symbols script. This was useful for testing but adds a lot of size to
this PR. Open to other ideas on where to place this. A separate test fixtures repo? Github LFS?

### DIFF
--- a/tests/integration/size/test_strip_symbols_insight.py
+++ b/tests/integration/size/test_strip_symbols_insight.py
@@ -1,0 +1,34 @@
+from pathlib import Path
+
+from launchpad.artifacts.apple.zipped_xcarchive import ZippedXCArchive
+from launchpad.size.analyzers.apple import AppleAppAnalyzer
+
+
+class TestStripSymbolsInsightIntegration:
+    def test_strip_symbols_with_hackernews_analyzer(self):
+        artifact_path = Path("tests/_fixtures/ios/HackerNews-Debug.xcarchive.zip")
+
+        analyzer = AppleAppAnalyzer(
+            skip_swift_metadata=True,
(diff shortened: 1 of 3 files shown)


tests/integration/size/test_strip_symbols_insight.py
```diff
@@ -0,0 +1,42 @@
...
+import pytest
+
+from launchpad.artifacts.apple.zipped_xcarchive import ZippedXCArchive
+from launchpad.size.analyzers.apple import AppleAppAnalyzer
+
+
+class TestStripSymbolsInsightIntegration:
+    def test_strip_symbols_with_hackernews_analyzer(self):
+        # Skip this test if we are not on a mac
+        if not platform.system() == "Darwin":
```
runningcode, line 14: Test is skipped on non-mac systems since it doesn't work.

Should I add a mac agent to run this test?

runningcode: The `T` option for the `strip` command is invalid.
https://github.com/getsentry/launchpad/actions/runs/16197471524/job/45727625524#step:7:287

From the man page, it seems this is important:
```
 The intent of this flag is to remove Swift symbols from the Mach-O symbol table, It removes the symbols whose names begin with `_$S' or `_$s' only when it finds an __objc_imageinfo section with and
              it has a non-zero swift version.  The future the implementation of this flag may change to match the intent.  When used together with -R,/ -s files the Swift symbols will also be removed from global
              symbol lists used by dyld.
```

PR #168, opened 2025-07-10
Add simulated app thinning

trevor-e:
This introduces some basic app thinning for duplicate image resources + architectures. Since we are running on Linux, we don't have access to `assetutil` for true thinning.

Now we at least match our old treemap:
Screenshot 2025-07-10 at 6 59 37 PM


src/launchpad/size/analyzers/apple.py
```diff
@@ -105,6 +106,10 @@ def analyze(self, artifact: AppleArtifact) -> AppleAnalysisResults:
 
         file_analysis = self._analyze_files(artifact)
         logger.info(f"Found {file_analysis.file_count} files, total size: {file_analysis.total_size} bytes")
+        file_analysis = AppThinningSimulator.for_iphone_se().apply_thinning(file_analysis)
```
noahsmartin, line 109: Our old thinning did 3x isn't that what we should do here?

trevor-e:
Change of approach: we're going to require users perform their own app thinning before uploading.

Issue #98, opened 2025-06-30
Running docker tests locally


chromy:
Docker tests now succeed in the CI but fail when ran locally on a MacOS arm64 laptop.

Setup as normal, then:
```
docker build --build-arg TEST_BUILD=true -t launchpad-test .
```

Run the tests:
```
docker run  \
    -e LAUNCHPAD_ENV=development \
    -e LAUNCHPAD_HOST=localhost \
    -e LAUNCHPAD_PORT=2218 \
    -e KAFKA_BOOTSTRAP_SERVERS="localhost:9092" \
    -e KAFKA_GROUP_ID="launchpad-test" \
    -e KAFKA_TOPICS="preprod-artifact-events" \
    launchpad-test python -m pytest tests/ -v
```

Some tests which use `bundletool` fail. 

Running `bundletool` manually:
```
docker run launchpad-test bash -c 'bundletool build-apks --bundle /app/tests/_fixtures/android/hn.aab --output foo.apks'
```

Lots of errors of the form:
```
Jun 30, 2025 5:05:31 PM com.google.common.util.concurrent.AggregateFuture log
SEVERE: Got more than one input Future failure. Logging failures after the first
com.android.tools.build.bundletool.model.exceptions.CommandExecutionException: Command '[/tmp/AutoValue_BuildApksCommand17763852638748728593/output/linux/aapt2, convert, --output-format, binary, -o, /tmp/14906132183331271825/binary156.apk, /tmp/14906132183331271825/proto155.apk]' didn't terminate successfully (exit code: 133). Check the logs.
        at com.android.tools.build.bundletool.model.exceptions.InternalExceptionBuilder.build(InternalExceptionBuilder.java:57)
        at com.android.tools.build.bundletool.androidtools.DefaultCommandExecutor.executeImpl(DefaultCommandExecutor.java:71)
        at com.android.tools.build.bundletool.androidtools.DefaultCommandExecutor.execute(DefaultCommandExecutor.java:38)
        at com.android.tools.build.bundletool.androidtools.Aapt2Command$1.convertApkProtoToBinary(Aapt2Command.java:71)
        at com.android.tools.build.bundletool.io.Aapt2ResourceConverter$ResourceConverter.convertAndOptimizeProtoApk(Aapt2ResourceConverter.java:158)
        at com.android.tools.build.bundletool.io.Aapt2ResourceConverter$ResourceConverter.convertResourcesToBinary(Aapt2ResourceConverter.java:125)
[...]

PR #176, opened 2025-07-12
Move external deps download step into dockerfile (docker build)

NicoHinderling:
When testing locally, I noticed certain unit tests were failing because cwldemangler wasn't actually being installed (more: https://github.com/getsentry/launchpad/issues/98)

Previously the way this was intended to work is:

- You run scripts/deps in the build directory which downloads all the dependencies for both darwin-aarch64 and linux-x86_64 to $BUILD_DIR/.devenv/deps
- You run docker build ... to build the image
- In the Dockerfile we COPY the relevant binaries from $BUILD_DIR/.devenv/deps into /usr/local/bin/ of the Docker image.

Now we do the scripts/deps download as part of the docker build step

### DIFF
--- a/scripts/deps
+++ b/scripts/deps
@@ -100,7 +100,6 @@ class Dep:
 
     def matches_platform(self, architecture, system):
         """Check if this dep matches the given platform."""
-        # Platform-independent deps match all platforms
         if not self.is_machine_dependent():
             return True
         return self.architecture == architecture and self.system == system
@@ -290,8 +289,6 @@ def remove_tree(path):
 def install_to_system(deps, architecture, system):
     """Install binaries from deps to /usr/local/bin/"""
     install_dir = "/usr/local/bin"
-
-    # Check if we have write permission
     if not os.access(install_dir, os.W_OK):
         print(f"Error: No write permission to {install_dir}. You may need to run with sudo.")
         return False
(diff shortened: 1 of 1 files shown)


scripts/deps
```diff
@@ -98,6 +98,13 @@ class Dep:
     def get_binaries(self):
         return self.binaries or []
 
+    def matches_platform(self, architecture, system):
+        """Check if this dep matches the given platform."""
+        # Platform-independent deps match all platforms
```
chromy, line 103: nit: this comment probably not needed

scripts/deps
```diff
@@ -273,6 +287,46 @@ def remove_tree(path):
         os.remove(path)
 
 
+def install_to_system(deps, architecture, system):
+    """Install binaries from deps to /usr/local/bin/"""
+    install_dir = "/usr/local/bin"
+
+    # Check if we have write permission
```
chromy, line 294: nit: not needed

scripts/deps
```diff
@@ -273,6 +287,46 @@ def remove_tree(path):
...
+
+    installed = []
+    for dep in deps:
+        if not dep.matches_platform(architecture, system):
+            continue
+
+        for binary in dep.get_binaries():
+            source_path = dep.get_binary_target_path(binary)
+            if not os.path.exists(source_path):
+                print(f"Warning: {source_path} not found, skipping")
```
chromy, line 307: This one seems like it should be an error, if we can't install something we probably need to take some action to fix it.

scripts/deps
```diff
@@ -273,6 +287,46 @@ def remove_tree(path):
...
+        print(f"Error: No write permission to {install_dir}. You may need to run with sudo.")
+        return False
+
+    installed = []
+    for dep in deps:
+        if not dep.matches_platform(architecture, system):
+            continue
+
+        for binary in dep.get_binaries():
+            source_path = dep.get_binary_target_path(binary)
```
chromy, line 305: `deps.get_binary_source_path` maybe is better. It's the same in the end but currently this reads like a typo (`source_path = dep.get_binary_target_path(binary)`).

chromy (approved):
lgtm! % some small comments.

Could you also update the PR title and description to explain the new way this works before landing. 🙏🏻

PR #179, opened 2025-07-14
Add sentry SDK to launchpad service

NicoHinderling:
allows us to monitor the service in production

### DIFF
--- a/src/launchpad/sentry_sdk_init.py
+++ b/src/launchpad/sentry_sdk_init.py
@@ -24,7 +24,7 @@ def initialize_sentry_sdk() -> None:
     """Initialize Sentry SDK with launchpad-specific configuration."""
     config = get_sentry_config()
 
-    if config.get("environment") in ("TEST", "DEVELOPMENT"):
+    if config.get("environment", "").lower() in ("test", "development"):
         logger.debug(f"In {config.get('environment')} environment, skipping Sentry SDK initialization")
         return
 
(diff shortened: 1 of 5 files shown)


src/launchpad/sentry_sdk_init.py
```diff
@@ -0,0 +1,90 @@
...
+    if not config.get("dsn"):
+        logger.info("Sentry DSN not provided, skipping Sentry SDK initialization")
+        return
+
+    def before_send(event: Event, hint: dict[str, Any]) -> Event | None:
+        """Filter out internal errors from Sentry reporting."""
+        if "exc_info" in hint:
+            exc_type, exc_value, tb = hint["exc_info"]
+            # Exclude errors intended for internal handling, not Sentry
+            if isinstance(exc_value, LaunchpadInternalError):
```
trevor-e, line 49: What's the purpose of this, are there types of errors we don't want to report? I don't see LaunchpadInternalError used anywhere yet. Would be nice to have an example of what you intend to do with this b/c I can't think of any right now.

src/launchpad/sentry_sdk_init.py
```diff
@@ -0,0 +1,90 @@
...
+
+    pass
+
+
+def initialize_sentry_sdk() -> None:
+    """Initialize Sentry SDK with launchpad-specific configuration."""
+    config = get_sentry_config()
+
+    # Only initialize Sentry SDK in production environment
+    if config.get("environment") != "production":
```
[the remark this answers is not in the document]
trevor-e: Why wouldn't we want it running during tests/localdev in a separate environment? I've looked at a few dev errors in the past for other projects.

src/launchpad/sentry_sdk_init.py
NicoHinderling: yeah frankly i can't either. I added it pre-emptively. I'll just remove it for now

PR #183, opened 2025-07-15
Add proguard mapping support

rbro112:


src/launchpad/artifacts/android/aab.py
```diff
@@ -75,9 +77,25 @@ def get_primary_apks(self, device_spec: DeviceSpec = DeviceSpec()) -> list[APK]:
...
         finally:
             cleanup_directory(apks_dir)
+
+    def get_dex_mapping(self) -> DexMapping | None:
+        if self._dex_mapping is not None:
+            return self._dex_mapping
+
+        dex_mapping_files = list(self._extract_dir.rglob("proguard.map"))
+        if len(dex_mapping_files) > 1:
+            raise ValueError("Multiple dex_mapping.txt files found in AAB")
```
trevor-e, line 93: One thing I've been meaning to do at some point is create error subclasses instead of using ValueError everywhere, I believe that's the idiomatic way similar to Ruby.

src/launchpad/parsers/android/dex/dex_mapping.py
```diff
@@ -0,0 +1,242 @@
...
+            if line.startswith("#"):
+                pending_file_name = self._parse_comment(line, pending_file_name)
+            elif line.startswith(" "):
+                trimmed = line.strip()
+                if trimmed.startswith("#"):
+                    pending_file_name = self._parse_comment(line, pending_file_name)
+                else:
+                    current_class = self._parse_method_or_member(line, current_class)
+            elif line.endswith(":"):
+                current_class, pending_file_name = self._parse_class(line, current_class, pending_file_name, classes)
```
trevor-e, line 41: Is it worth having an else clause here or are we only parsing a subset of the file?

src/launchpad/parsers/android/dex/dex_mapping.py
```diff
@@ -0,0 +1,242 @@
...
+
+        Returns:
+            The deobfuscated field name, or None if not found
+        """
+        # First try to find the class by obfuscated name
+        clazz = self.lookup_obfuscated_class(class_name)
+        if clazz is None:
+            # Try to find by deobfuscated signature
+            clazz = self.lookup_deobfuscated_signature(class_name)
+            if clazz is None:
```
trevor-e, line 230: nit: could these branches be flattened? if i'm reading correctly they don't need to be nested.

trevor-e (approved):

src/launchpad/parsers/android/dex/dex_mapping.py
rbro112: I don't see how it can, basically these are "fallbacks" if we can't find the deobfuscated class_name, each one is further trying to get us the proper deobfuscated class if the previous step failed

src/launchpad/artifacts/android/aab.py
rbro112: We can add a follow up to do better error auditing. I think we should think through best practice here and move forward once we decide on the best plan.

PR #191, opened 2025-07-16
Add loose image insights

trevor-e:


src/launchpad/size/insights/apple/loose_images.py
```diff
@@ -0,0 +1,147 @@
...
+        match = self.CANONICAL_NAME_PATTERN.match(filename)
+        if match:
+            base_name = match.group(1)
+            extension = match.group(2)
+            return base_name + extension
+
+        return filename  # Fallback to original filename if pattern doesn't match
+
+    def _get_eliminated_files(self, group: LooseImageGroup) -> list[str]:
+        """Get list of file paths that would be eliminated via app thinning for @3x devices."""
```
trevor-e, line 113: I picked 3x since that would give the biggest app thinning savings (getting rid of 2x + 1x images)

src/launchpad/size/insights/apple/loose_images.py
```diff
@@ -0,0 +1,147 @@
...
+            return False
+
+        # Skip .stickerpack directories (as mentioned in Swift code)
+        path_parts = file_info.path.split("/")
+        if any(part.endswith(".stickerpack") for part in path_parts):
+            return False
+
+        return True
+
+    def _get_canonical_image_name(self, file_path: str) -> str:
```
trevor-e, line 100: This captures more than our old insight but I think it seems right after testing on some apps. The old logic required `@` in the URL.

src/launchpad/size/insights/apple/loose_images.py
```diff
@@ -0,0 +1,147 @@
...
+        """Extract canonical image name by removing resolution suffixes like @2x, @3x, ~ipad."""
+        filename = file_path.split("/")[-1]  # Get just the filename
```
noahsmartin, line 102: maybe better to use basename?

src/launchpad/size/insights/apple/loose_images.py
```diff
@@ -0,0 +1,147 @@
...
+        image_groups = [
+            LooseImageGroup(canonical_name=canonical_name, images=images)
+            for canonical_name, images in image_groups_dict.items()
+        ]
+
+        image_groups.sort(key=lambda group: group.total_size, reverse=True)
+
+        # Calculate total savings and avoid double-counting:
+        # 1. Files eliminated via app thinning: full block-aligned disk usage saved
+        # 2. Files that remain: only block alignment waste saved
```
noahsmartin, line 52: technically there is also the code signing overhead. Every new file creates a hew hash in the CodeResources file. Maybe fine to skip that here, it would only make the savings higher

src/launchpad/size/insights/apple/loose_images.py
```diff
@@ -0,0 +1,147 @@
...
+        )
+
+    def _is_loose_image_file(self, file_info: FileInfo) -> bool:
+        """Check if a file is a raw image that should be moved to an asset catalog."""
+        # Must be an image file
+        if file_info.file_type not in self.IMAGE_EXTENSIONS:
+            return False
+
+        # Skip if it's already in an asset catalog (.car file)
+        if ".car/" in file_info.path:
```
noahsmartin, line 85: A .car isn't a folder, I don't think we should have this check

src/launchpad/size/insights/apple/loose_images.py
trevor-e: I don't think our old code handled that either but I'll leave a TODO since that would be nice to include.

PR #192, opened 2025-07-16
Add app_id as common between Android and iOS, wireup frontend to use it

rbro112:
Set our universal `app_id` terminology rather than use separate `bundle_id` and `package_name` on iOS

Screenshot 2025-07-16 at 1 41 33 PM


NicoHinderling (approved):

NicoHinderling:
i wonder if there's a strong chance we'll need android specific fields in the future, but will defer to ryan

PR #198, opened 2025-07-17
Add iOS image optimization insight

trevor-e:


src/launchpad/artifacts/apple/zipped_xcarchive.py
```diff
@@ -23,11 +23,12 @@
...
 class AssetCatalogElement:
     name: str
-    imageId: str
+    image_id: str
     size: int
-    type: str
+    type: int
     vector: bool
     filename: str
+    full_path: Path
```
trevor-e, line 31: I couldn't find a better spot to store this.

src/launchpad/size/insights/apple/image_optimization.py
```diff
@@ -0,0 +1,175 @@
...
+    TARGET_HEIC_QUALITY = 85
+    _MAX_WORKERS = 8
+
+    def generate(self, input: InsightsInput) -> ImageOptimizationInsightResult | None:  # noqa: D401
+        files = list(self._iter_optimizable_files(input.file_analysis.files))
+        if not files:
+            return None
+
+        results: List[OptimizableImageFile] = []
+        with ThreadPoolExecutor(max_workers=min(self._MAX_WORKERS, len(files))) as executor:
```
trevor-e, line 53: This was slower than I hoped so I threw it in a threadpool. I'll monitor the performance (e.g. peak memory) of this.

src/launchpad/size/insights/apple/image_optimization.py
```diff
@@ -0,0 +1,175 @@
...
+                new_size = buf.tell()
+            return _OptimizationResult(file_size - new_size, new_size) if new_size < file_size else None
+        except Exception as exc:
+            logger.error("HEIC minification check failed: %s", exc)
+            return None
+
+    def _iter_optimizable_files(self, files: Sequence[FileInfo]) -> Iterable[FileInfo]:
+        for fi in files:
+            if fi.file_type == "car":
+                yield from (c for c in fi.children if self._is_optimizable_image_file(c))
```
runningcode, line 165: oh, this is a cool syntax for going through the car file. I didn't know you could do that! very cool.

runningcode (approved):
learned some nice syntax from reviewing this!

src/launchpad/size/insights/apple/image_optimization.py
trevor-e: Yea this is called a "generator" function since it yields values to iterators. In this case it does make the code a lot nicer to read since it handles the nested-ness of `car` files without having to modify the calling code.

src/launchpad/artifacts/apple/zipped_xcarchive.py
trevor-e: Do you remember which ones? Something still feels organizationally off so might do a refactoring down the road for this.

noahsmartin: should this be optional? Some asset catalog elements will be things like colors or NSDataAsset that will not have a corresponding file in the "ParsedAssets" directory

noahsmartin: I also wonder if this should be a function since it can be entirely derived from the other existing fields

src/launchpad/size/analyzers/apple.py
```diff
@@ -383,29 +386,29 @@ def _analyze_files(self, xcarchive: ZippedXCArchive) -> FileAnalysis:
...
+        result: List[FileInfo] = []
+        for element in catalog_details:
+            if element.full_path.exists() and element.full_path.is_file():
+                file_hash = calculate_file_hash(element.full_path, algorithm="md5")
+            else:
+                # can't hash a directory so use imageId as hash
+                file_hash = element.image_id
+
+            result.append(
+                FileInfo(
```
noahsmartin, line 401: Is it going to cause problems that images are treated as "files" when they aren't actually files? Like will this break the insight about how many files are in the app, or the calculation of code signature size that increases as you have more files? Wasn't sure if any of those use this variable

src/launchpad/artifacts/apple/zipped_xcarchive.py
runningcode: I added `full_path` to `FileInfo` with the `exclude=true` as you suggested.

src/launchpad/size/analyzers/apple.py
trevor-e: Seems like it correctly ignores the nested children.

PR #200, opened 2025-07-18
Use uv provided by devenv

chromy:
As of https://github.com/getsentry/devenv/pull/198 (1.22.0)
devenv includes support for `uv`. This switches us to use that version
of 'uv'. It doesn't yet unleash the full power of `uv` (uv sync,
uv lock, etc) but it cuts down on the amount of custom bits we have.

If you get an error like:

ImportError: cannot import name 'uv' from 'devenv.lib'

You need to update your version of devenv. Run:

rm -rf .venv
devenv update
devenv sync


chromy:
cc: @romtsn

PR #210, opened 2025-07-22
Add static inits detection

trevor-e:


src/launchpad/parsers/apple/macho_parser.py
```diff
@@ -236,3 +241,39 @@ def has_swift_imageinfo(self) -> bool:
...
             logger.error("Could not parse __objc_imageinfo: %s", e)
             return False
+
+    @trace(name="static_inits")
+    def static_inits(self) -> List[lief.Symbol | str]:
+        init_sec = self.get_section_bytes("__mod_init_func")
+        if not init_sec:
+            return []
+
+        ptr_size = 8 if self.binary.header.cpu_type == lief.MachO.Header.CPU_TYPE.ARM64 else 4
```
trevor-e, line 251: This feels wrong but don't know the correct solutions right now. Do we need this check? And is the pointer size based on CPU or just whether it's 32/64 bit?

src/launchpad/parsers/apple/macho_parser.py
```diff
@@ -236,3 +241,39 @@ def has_swift_imageinfo(self) -> bool:
...
+        endianness = "<"  # Mach-O on x86-64/arm64 is little-endian
+        fmt = f"{endianness}{'Q' if ptr_size == 8 else 'I'}"
+
+        addrs = [struct.unpack(fmt, init_sec[i : i + ptr_size])[0] for i in range(0, len(init_sec), ptr_size)]
+
+        symbols_by_addr = sorted((s for s in self.binary.symbols if s.value), key=lambda s: s.value)
```
noahsmartin, line 257: Will this use the dsym? Or will it only work for binaries that don't strip their symbols?

src/launchpad/parsers/apple/macho_parser.py
trevor-e: good idea

PR #212, opened 2025-07-23
Fix various insights

trevor-e:
Testing against some existing client apps and found a number of discrepancies that should be fixed now:

- fixes FileAnalysis to store the file size adjusted for the block size, this was causing a number of insight savings to be wrong
- fixes Loose Images to only report images that have at least one scale
- fixes binary export metadata to only report if over 1024 byte savings
- fixes various insights to return None when there is no data
- fixes the Duplicate Files insight to group by the filename so it's easier to display in the frontend
- adds a new insights UI for debugging
- use relative paths for binary strip results

### DIFF
--- a/src/launchpad/size/insights/common/duplicate_files.py
+++ b/src/launchpad/size/insights/common/duplicate_files.py
@@ -32,12 +32,12 @@ class DuplicateFilesInsight(Insight[DuplicateFilesInsightResult]):
                     group = DuplicateFileGroup(
                         filename=group_filename,
                         files=sorted_files,
-                        savings=savings_for_this_group,
+                        total_savings=savings_for_this_group,
                     )
                     groups.append(group)
                     total_savings += savings_for_this_group
 
-        groups = sorted(groups, key=lambda g: (-g.savings, g.filename))
+        groups = sorted(groups, key=lambda g: (-g.total_savings, g.filename))
 
         return DuplicateFilesInsightResult(
             groups=groups,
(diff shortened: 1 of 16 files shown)


src/launchpad/size/insights/common/duplicate_files.py
```diff
@@ -13,20 +15,31 @@ def generate(self, input: InsightsInput) -> DuplicateFilesInsightResult:
             if file.hash_md5:
                 files_by_hash[file.hash_md5].append(file)
 
-        duplicate_files: List[FileInfo] = []
+        groups: List[DuplicateFileGroup] = []
         total_savings = 0
 
         for file_list in files_by_hash.values():
             if len(file_list) > 1:
+                # Calculate savings: total size - size of one copy we keep
```
runningcode, line 23: that makes more sense! thanks for fixing

src/launchpad/size/models/apple.py
```diff
@@ -71,6 +58,29 @@ class SmallFilesInsightResult(BaseInsightResult):
...
+
+    @computed_field
+    @property
+    def total_savings(self) -> int:
+        """Size savings by moving to asset catalog (excluding largest image that would be kept)."""
+        if len(self.images) <= 1:
+            return 0
+
+        # Sort by size descending, exclude the largest (highest scale) image
+        sorted_images = sorted(self.images, key=lambda img: img.size, reverse=True)
```
noahsmartin, line 80: would be technically faster to just create the array of sizes, find the max, and subtract it from the total sum. Then you wouldn't have to do the Nlog(N) sort

PR #216, opened 2025-07-24
Another batch of insight fixes

trevor-e:
- Fixes bug where Watch/AppExtension binaries weren't considered for the export metadata insight
- Fixes the duplicate files insight to group the results by file, previously everything was flat. Also fixes some edge cases where it would report duplicates within nested folders.
  - Added special grouping for `.bundle` folders and allowlisted `.xcprivacy` files
- Introduces a new `FileSavingsResult` to share amongst insights which is more slimmed down and should make the response JSON smaller.

### DIFF
--- a/src/launchpad/size/insights/common/duplicate_files.py
+++ b/src/launchpad/size/insights/common/duplicate_files.py
@@ -1,12 +1,11 @@
-import hashlib
 import os
 
 from collections import defaultdict
-from pathlib import Path
-from typing import Dict, List
+from pathlib import PurePosixPath
+from typing import Dict, List, Set
 
 from launchpad.size.insights.insight import Insight, InsightsInput
-from launchpad.size.models.common import FileInfo, TreemapType
+from launchpad.size.models.common import FileInfo
 from launchpad.size.models.insights import (
     DuplicateFileGroup,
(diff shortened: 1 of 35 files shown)


src/launchpad/size/insights/common/duplicate_files.py
```diff
@@ -1,45 +1,126 @@
...
 class DuplicateFilesInsight(Insight[DuplicateFilesInsightResult]):
-    def generate(self, input: InsightsInput) -> DuplicateFilesInsightResult:
-        files_by_hash: Dict[str, List[FileInfo]] = defaultdict(list)
-        for file in input.file_analysis.files:
-            if file.hash_md5:
-                files_by_hash[file.hash_md5].append(file)
+    EXTENSION_ALLOWLIST = [".xcprivacy"]
+
+    # Make sure to group all duplicates in a directory that has one of these extensions
+    DIRECTORY_EXTENSIONS = [".bundle"]
```
noahsmartin, line 20: The emerge logic would group any directory right? Like IIRC react native apps commonly have a bunch of folders, and we would group duplicates in folders even if those folders were not bundles

noahsmartin (approved):

PR #222, opened 2025-07-25
Use pre-built Linux-compatible strip tool

trevor-e:
I found a project that handles porting some Darwin tools to Linux: https://github.com/tpoechtrager/cctools-port

This adds some scripts to compile that project into binaries, and then replaces our current implementation to use the binary approach. Unfortunately our old approach using `lief` wasn't giving viable results. From what I can tell in testing, this gives me the exact same result on Linux that I see locally on my MBP.

**Follow-up**: right now the binaries are hard-coded in the PR. I'll work with @chromy on how to get these stored in a better way to work with `deps`.


Dockerfile
```diff
@@ -1,3 +1,24 @@
+# Build libdispatch for the strip binary
+FROM --platform=linux/amd64 debian:12-slim AS libdispatch-build
```
trevor-e, line 2: These changes install a bunch of runtime dependencies needed by `strip` and `ld`

scripts/strip/build-strip-linux.sh
trevor-e, line 1: We'll probably very rarely (if every) build this binary again so I just stuffed the build script in here for now.

scripts/strip/dist/ld
trevor-e, line 1: I'll replace these with our `deps` handling once I figure out how we want to store them. Do we have a GCP bucket? Would that break if someone tried to fork this project?

trevor-e:
@sentry review

runningcode (approved):

.github/workflows/ci.yml
```diff
@@ -106,11 +106,21 @@ jobs:
...
+          docker run --rm \
+            -e LAUNCHPAD_ENV=development \
+            -e LAUNCHPAD_HOST=localhost \
+            -e LAUNCHPAD_PORT=2218 \
+            -e KAFKA_BOOTSTRAP_SERVERS="localhost:9092" \
+            -e KAFKA_GROUP_ID="launchpad-test" \
+            -e KAFKA_TOPICS="preprod-artifact-events" \
+            -v ${{ github.workspace }}:/workspace \
+            -w /workspace \
+            launchpad-test python -m pytest tests/unit/ tests/integration/ -v --cov --cov-branch --cov-report=xml --junitxml=junit.xml
```
trevor-e, line 123: This happened to work before but now also needs to run in Docker.

Dockerfile
```diff
@@ -44,12 +76,23 @@ COPY pyproject.toml .
 COPY README.md .
 COPY LICENSE .
 
+# Copy libdispatch from the build stage
+COPY --from=libdispatch-build /usr/lib/x86_64-linux-gnu/libdispatch.so* /usr/lib/x86_64-linux-gnu/
+
+# Copy and verify the strip and ld binaries, then make them executable
+COPY scripts/strip/dist/strip scripts/strip/dist/ld /app/scripts/strip/dist/
+RUN echo "4cd01dd28294a3ebeff031d6ba947aee1c2dd9c402f504f9866eec302466b11d  /app/scripts/strip/dist/strip" | sha256sum -c - && \
```
trevor-e, line 84: Added checksum verification for the binaries, I think our `deps` does this too once I migrate over to using that.

noahsmartin (approved):

.github/workflows/ci.yml
chromy: I made it work before on purpose, I think it would be nice if we don't have to run tests in Docker because I think that puts another step in the way of debugging test failures when they come up. However I guess there is a balance so if gets to the point we have to use docker I because we have to many weird tools thats reasonable but I think it should be a decision we take rather than the non-docker build just gradually eroding.

PR #227, opened 2025-07-28
[WIP] Clean up insights schema

trevor-e:


src/launchpad/size/models/insights.py
```diff
@@ -14,70 +20,188 @@ class FileSavingsResult(BaseModel):
...
-    """Base class for all insight results."""
+class FileSavingsResultGroup(BaseModel):
+    """Group of files with savings information."""
 
     model_config = ConfigDict(frozen=True)
 
-    total_savings: int = Field(..., ge=0, description="Total potential savings in bytes")
+    name: str = Field(..., description="Group name or identifier")
+    files: List[FileSavingsResult] = Field(..., description="Files in this group")
+    total_savings: int = Field(..., ge=0, description="Total savings for this group")
```
rbro112, line 30: Should we just rely on the sum of each `file` object for total_savings, or similar to the treemap would it be best to have this field?

Additionally, download vs install savings could be good to separate here and in `FileSavingsResult`

src/launchpad/size/models/insights.py
```diff
@@ -2,7 +2,13 @@
...
 from pydantic import BaseModel, ConfigDict, Field
 
-from .common import FileInfo
+
+class BaseInsightResult(BaseModel):
+    """Base class for all insight results."""
+
+    model_config = ConfigDict(frozen=True)
+
+    total_savings: int = Field(..., ge=0, description="Total potential savings in bytes")
```
rbro112, line 11: Separate download vs install?

rbro112 (approved):

PR #232, opened 2025-07-29
Remove granular notion of download vs install size

trevor-e:


trevor-e:
Ryan already snagged 😢

PR #255, opened 2025-08-01
iOS treemap improvements

trevor-e:
Fixes several treemap bugs/missing features for iOS (like segment grouping) and deleted a lot of code. The algorithm now is a lot simpler and has no more indirection. We simply keep track of all the symbol sizes to subtract from the respective sections/segments, and then do a loop through to add all the segments/sections that still have leftover size to attribute. This also fixes the missing `LINKEDIT` segment.

For HackerNews I believe everything is matching up correctly. For some customer apps I'm occasionally seeing some binaries overcounting bytes by ~1%, still looking into that.

Screenshot 2025-08-04 at 4 52 47 PM


src/launchpad/size/models/apple.py
```diff
@@ -72,32 +71,67 @@ class AppleAppInfo(BaseAppInfo):
     main_binary_uuid: str | None = Field(None, description="UUID of the main binary")
 
 
-class MachOBinaryAnalysis(BaseBinaryAnalysis):
-    """Mach-O binary analysis results."""
+@dataclass
+class SegmentInfo:
```
trevor-e, line 75: I created types to capture some of the `lief` info we were having issues with memory leaks using data after the file is closed.

src/launchpad/size/models/binary_component.py
```diff
@@ -52,72 +49,7 @@ class BinaryTag(Enum):
     UNWIND_INFO = "unwind_info"
     DEBUG_INFO = "debug_info"
 
+    OTHER = "other"
+
     # Unmapped regions
     UNMAPPED = "unmapped"
-
-
-class BinaryComponent(BaseModel):
```
trevor-e, line 59: no longer need

src/launchpad/size/analyzers/apple.py
```diff
@@ -472,3 +473,46 @@ def _check_strip_symbols_removal(self, binary_path: Path, binary: lief.MachO.Bin
...
+                    segment_name = command.name
+                    if isinstance(segment_name, bytes):
+                        segment_name = segment_name.decode("utf-8", errors="replace")
+                    else:
+                        segment_name = str(segment_name)
+
+                    section_infos: List[SectionInfo] = []
+                    for section in command.sections:
+                        section_name = section.name
+                        if isinstance(section_name, bytes):
```
noahsmartin, line 493: could dedup this and the segment name handling with a function that does the if statement

src/launchpad/size/treemap/macho_element_builder.py
```diff
@@ -225,81 +236,239 @@ def _tree_to_treemap(node_map: Dict[str, _SwiftTypeNode]) -> List[TreemapElement
...
+                            name=section_name,
+                            size=adjusted,
+                            type=self._get_element_type_from_tag(tag),
+                            path=None,
+                            is_dir=False,
+                            children=[],
+                        )
+                    )
+
+            if segment_name == "__LINKEDIT":
```
noahsmartin, line 297: We maybe should handle any segment that has no sections, rather than specifically hardcoding linkedit

src/launchpad/size/treemap/macho_element_builder.py
```diff
@@ -225,81 +236,239 @@ def _tree_to_treemap(node_map: Dict[str, _SwiftTypeNode]) -> List[TreemapElement
...
+                    )
+                )
+
+        # Add an explicit "Unmapped" region if present (simplified - just check if we have unaccounted bytes)
+        total_accounted = sum(c.size for c in binary_children)
+        if binary_analysis.executable_size > total_accounted:
+            unaccounted = binary_analysis.executable_size - total_accounted
+            binary_children.append(
+                TreemapElement(
+                    name="Unmapped",
```
noahsmartin, line 321: So the emerge version had "Unmapped" for every segment, this just does one unmapped for all segments grouped together, right? I think that's fine, just confirming

PR #269, opened 2025-08-06
Enable deploy for US region

NicoHinderling:


chromy (approved):
lgtm but we might want to wait until the ops side is in before landing so as not to get gocd failures

PR #281, opened 2025-08-08
SmallStrings insight fixes

trevor-e:
- exclude `LaunchScreen.strings` since I don't believe that's usable in code
- remove the files list from the response, we don't want to duplicate this with our other minify script
- only show the insight if the estimated savings is greater than 100KB

### DIFF
--- a/src/launchpad/size/insights/apple/localized_strings.py
+++ b/src/launchpad/size/insights/apple/localized_strings.py
@@ -1,13 +1,24 @@
+import re
+
+from pathlib import Path
+
 from launchpad.size.insights.insight import Insight, InsightsInput
 from launchpad.size.models.common import FileInfo
-from launchpad.size.models.insights import FileSavingsResult, LocalizedStringInsightResult
+from launchpad.size.models.insights import LocalizedStringInsightResult
 
 
 class LocalizedStringsInsight(Insight[LocalizedStringInsightResult]):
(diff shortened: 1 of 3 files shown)


src/launchpad/size/insights/apple/localized_strings.py
```diff
@@ -18,19 +29,18 @@ def generate(self, input: InsightsInput) -> LocalizedStringInsightResult | None:
...
-            file_savings = [FileSavingsResult(file_path=file.path, total_savings=file.size) for file in localized_files]
-
+            if self._LPROJ_STRINGS_PATTERN.search(file_info.path):
+                filename = Path(file_info.path).name
+                if filename not in self.STRINGS_FILE_DENYLIST:
+                    localized_files.append(file_info)
+                    total_size += file_info.size
+
+        estimated_savings = int(total_size * self.ESTIMATED_SAVINGS_RATIO)
+        if estimated_savings > self.THRESHOLD_BYTES:
```
noahsmartin, line 41: I would suggest changing this to be something like `if total_size > (100 * 1024)/0.8` and not calculate an "estimated_savings" since our insights tend to be more precise and not estimates. Instead the UI could just show an insight that you could see benefits from using this only when localized strings files are this large, that way we avoid any unsubstantiated hand wavy numbers

PR #282, opened 2025-08-13
Prepare for reinstating the health check

chromy:
Prepare for reinstating the health check.

- Use asyncio safe signal handlers
- Introduce a new wrapper class (`LaunchpadKafkaConsumer`) around the Kafka processor
- Move all handling of `health_check_file` into this wrapper
- Give `LaunchpadServer` and `LaunchpadKafkaConsumer` a consistent interface
   - `async start()` - start the component, completes as soon as the service is started
   - `async stop(timeout_s)` - stop the component, with a timeout of `timeout_s` seconds
   - `is_healthy() -> bool` - return true iff component is healthy
 - This consistent interface and moving `health_check_file` allows simplifying `LaunchpadService`
- Shutdown all components in parallel
-  Make healthcheck setup non-optional for `LaunchpadStrategyFactory`
 - Remove `HealthCheckResponse` for now and use a bool

### DIFF
--- a/src/launchpad/kafka.py
+++ b/src/launchpad/kafka.py
@@ -2,7 +2,9 @@
 
 from __future__ import annotations
 
+import asyncio
 import os
+import time
 
 from typing import Any, Callable, Dict, Mapping
 
@@ -20,7 +22,7 @@ from sentry_kafka_schemas.schema_types.preprod_artifact_events_v1 import (
     PreprodArtifactEvents,
 )
 
-from launchpad.constants import PREPROD_ARTIFACT_EVENTS_TOPIC
+from launchpad.constants import HEALTHCHECK_MAX_AGE_SECONDS, PREPROD_ARTIFACT_EVENTS_TOPIC
 from launchpad.utils.logging import get_logger
 
 logger = get_logger(__name__)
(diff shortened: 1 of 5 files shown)


src/launchpad/kafka.py
```diff
@@ -113,16 +176,12 @@ def create_with_partitions(
...
-            next_step = Healthcheck(self.healthcheck_file, next_step)
-            logger.info("KAFKASETUP: Healthcheck strategy added successfully")
-        else:
-            logger.warning("KAFKASETUP: No healthcheck file configured - skipping healthcheck strategy")
-            logger.info("KAFKASETUP: Processing will continue without healthcheck monitoring")
+        logger.info(f"KAFKASETUP: Healthcheck file configured: {self.healthcheck_file}")
+        logger.info("KAFKASETUP: Adding Healthcheck strategy to processing chain")
+        assert self.healthcheck_file
+        next_step = Healthcheck(self.healthcheck_file, next_step)
+        logger.info("KAFKASETUP: Healthcheck strategy added successfully")
```
NicoHinderling, line 184: im not wed to keeping these now that kafka appears to be stable but it doesn't fire too often so not a big deal

NicoHinderling (approved):

PR #286, opened 2025-08-13
Do not re-raise processing failures

chromy:


NicoHinderling (approved):

NicoHinderling (changes requested):
actually lets hold off on landing this for right now

PR #292, opened 2025-08-14
Delete two no longer used unit tests

NicoHinderling:


tests/integration/test_integration.py
```diff
@@ -17,41 +17,6 @@
 class TestServiceIntegration:
     """Integration tests for the full service."""
 
-    # @pytest.mark.asyncio
```
chromy, line 20: If would be good to have some test coverage of the health check. A test would have caught: https://github.com/getsentry/launchpad/pull/289 (type checker also should have caught that - but that is a problem for another day).

At minimum we could have a happy path test (`is_healthy()` returns true in an integration test for the {server,service,kafka}). Testing the failure case is harder but we could leave that for now.

chromy (approved):

PR #310, opened 2025-09-02
Reuse existing artifact already created in process_artifact

NicoHinderling:
ArtifactFactory.from_path reads the entire files into memory. We don't need to do it more than once

### DIFF
--- a/src/launchpad/service.py
+++ b/src/launchpad/service.py
@@ -168,6 +168,7 @@ class LaunchpadService:
 
         sentry_client = SentryClient(base_url=self._service_config.sentry_base_url)
         temp_file = None
+        artifact = None
 
         try:
             temp_file = self._download_artifact_to_temp_file(sentry_client, artifact_id, project_id, organization_id)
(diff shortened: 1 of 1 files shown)


src/launchpad/service.py
```diff
@@ -209,14 +210,15 @@ def process_artifact(self, artifact_id: str, project_id: str, organization_id: s
...
 
-            artifact = ArtifactFactory.from_path(file_path)
             if isinstance(artifact, ZippedXCArchive) and app_info.is_code_signature_valid and not app_info.is_simulator:
                 temp_dir = Path(tempfile.mkdtemp())
-                ipa_path = temp_dir / "App.ipa"
-                cast(ZippedXCArchive, artifact).generate_ipa(ipa_path)
-                sentry_client.upload_installable_app(organization_id, project_id, artifact_id, str(ipa_path))
-                self._safe_cleanup(str(ipa_path), "installable app")
-                logger.info(f"Successfully uploaded installable app for artifact {artifact_id}")
+                try:
```
trevor-e, line 215: nit: can this use the `with:` syntax?

trevor-e (approved):

PR #313, opened 2025-09-02
Fix dex parsing issue with character values

rbro112:
Dex parsing was hitting an issue after incorrectly parsing a character value. This fixes and confirms the fix against the offending build.

Resolves EME-228


trevor-e (approved):
Possible to add a test case?

PR #323, opened 2025-09-05
Only upload missing chunks

chromy:
This resolves two pending comments from:
9c40f95828d712b9009533462f9aec6f219d5377
- Putting back the 'retry' logic for assembling
- Using the new file based API rather than the path based one

In addition we refactor sentry_client.py to:
- Parse response types using Pydantic models
- Remove unused Result types
- Replace Dict[str, Any] in a number of return types
- Use responses for the remaining test_sentry_client tests
- Add a new SentryClientError for error handling replacing ErrorResult.
- Correctly handle the chunk options response - previously this was not
  working since we looked for snake_case args and the response was
  camelCase.
- Rename test_sentry_client_retry.py to tests/unit/test_sentry_client.py
  as it now contains tests for things beyond just retries.


tests/unit/test_sentry_client.py
```diff
@@ -0,0 +1,449 @@
...
+#
+#        result = client._make_json_request("POST", "/test", {"key": "value"})
+#
+#        assert result == {"result": "success"}
+#        assert mock_session.request.called
+#
+#        # Verify the call was made with correct parameters
+#        call_args = mock_session.request.call_args
+#        assert call_args[1]["method"] == "POST"
+#        assert "example.com/test" in call_args[1]["url"]
```
NicoHinderling, line 188: @chromy what's the plan with all of these?

PR #332, opened 2025-09-09
Add timing log for artifact download from monolith

NicoHinderling:
So we can see on average how long this takes


src/launchpad/kafka.py
```diff
@@ -39,7 +39,8 @@
 def create_kafka_consumer(
     message_handler: Callable[[PreprodArtifactEvents], Any],
 ) -> LaunchpadKafkaConsumer:
-    """Create and configure a Kafka consumer using environment variables."""
+    """Create and configure a Kafka consumer uspot
```
rbro112, line 42: Typo? What's `uspot sing`?

NicoHinderling: typo

PR #336, opened 2025-09-10
Pipe dequeued_at and processed_at back to monolith

NicoHinderling:
This involves calling the update endpoint at the very end. We will then pass these new fields in and have it saved to the `extra_info` column in the database.

### DIFF
--- a/src/launchpad/api/update_api_models.py
+++ b/src/launchpad/api/update_api_models.py
@@ -1,6 +1,7 @@
+from datetime import datetime
 from typing import List, Optional
 
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 
 class AppleAppInfo(BaseModel):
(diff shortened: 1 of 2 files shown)


src/launchpad/service.py
```diff
@@ -519,6 +526,28 @@ def _upload_results(
         else:
             logger.info(f"Successfully uploaded analysis results for artifact {artifact_id}")
 
+    def _send_final_update_with_timestamps(
```
trevor-e, line 529: Why do we need to make an API call at the end instead of just including the field in our JSON result we upload to the monolith? Like our `BaseAnalysisResults` already has an `analysis_duration` value in it, and are reading the file in `assemble_preprod_artifact_size_analysis`.

src/launchpad/api/update_api_models.py
```diff
@@ -15,11 +16,22 @@ class AppleAppInfo(BaseModel):
     # TODO: add "date_built" field once exposed in 'AppleAppInfo'
 
 
+class ExtraInfo(BaseModel):
```
trevor-e, line 19: I don't think this should be its own model, these fields being stored in the `extras` dict is an implementation detail on the model and not something we need to worry about here. This data can be inlined to `UpdateData` IMO.

src/launchpad/service.py
trevor-e: to be clear we do need to send the `dequeued_at` field as part of the app info request, but I don't think we need this second request.

PR #344, opened 2025-09-12
Temporary fix for AAB analysis

chromy:


chromy:
did a better fix in #357

PR #349, opened 2025-09-15
Create artifact context to clearly scope the lifecycle of the artifact

NicoHinderling:
Make sure that the lifetime and ownership is very clear for all artifact related content


NicoHinderling:
due to the layers of zips for android apps, I actually don't think this will be that great - just going to close it now

PR #350, opened 2025-09-15
Add ability to blocklist projects meant to be skipped

NicoHinderling:
If we ever are having an outage caused by a specific build / project and want to ignore them so that our kafka consumers can continue onto other messages, we can use this env variable and just do a k8s deploy to change the value


src/launchpad/service.py
```diff
@@ -527,6 +531,12 @@ def is_healthy(self) -> bool:
         is_kafka_healthy = self.kafka.is_healthy()
         return is_server_healthy and is_kafka_healthy
 
+    def _should_skip_processing_for_project(self, project_id: str) -> bool:
+        """Determine if processing should be skipped for a project."""
+
+        list_of_projects_to_skip = os.getenv("PROJECT_IDS_TO_SKIP", "").split(",")
```
chromy, line 537: should we put this in the ServiceConfig below?

tests/integration/test_service.py
```diff
@@ -128,3 +128,52 @@ def test_handle_kafka_message_error(self, mock_process):
...
+        service.handle_kafka_message(payload)
+
+        # Verify process_artifact was NOT called
+        mock_process.assert_not_called()
+
+        # Verify no metrics were recorded (since processing was skipped entirely)
+        calls = fake_statsd.calls
+        assert len(calls) == 0
+
+    @patch.dict("os.environ", {"PROJECT_IDS_TO_SKIP": "other-project"})
```
chromy, line 156: If we put PROJECT_IDS_TO_SKIP in the config we can avoid this patching which is nice.

chromy (approved):
lgtm % I think that reading the env variable should probably go in the config with the other parts

PR #354, opened 2025-09-16
Execute artifact processing in sub process

NicoHinderling:
I've been doing a fair amount of testing of the launchpad service. The android codepath seems to not have to substantial of memory leaks. The iOS one is pretty bad. 

image

image

I ran a bunch of memory profiling tests today using [memray](https://github.com/bloomberg/memray) (i might open another PR later to commit some code so we can easily & regularly memory profile going forward) and it seems like LEIF and pillow_leif are the main culprits. I tried implementing specific changes:

```
- Fix pillow_heif memory leak in image optimization - force cleanup of image buffers
- Fix LIEF memory leak in _extract_binary_uuid - explicitly delete parsed binaries
- Fix LIEF memory leak in _validate_executable - cleanup after validation
- Fix LIEF memory leak in _analyze_binary - cleanup parsed objects
- Clear PIL image cache after processing
```

but I only got down like 25% of the memory leaks. The goal is that if the service starts at 150mb, it should return to that after processing completes. I then came across subprocess isolation. This makes it very easy to ensure that our memory is totally cleaned up.

It appears to be working really well locally! I'd like to give it a shot in GCP and see how it goes! If this works, we will have a lot more headroom for memory usage over time in prod

### DIFF
--- a/src/launchpad/service.py
+++ b/src/launchpad/service.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 import json
+import multiprocessing
 import os
 import signal
 import tempfile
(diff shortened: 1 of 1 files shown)


src/launchpad/service.py
```diff
@@ -167,10 +187,33 @@ def handle_kafka_message(self, payload: PreprodArtifactEvents) -> None:
...
         """
         Download artifact and perform size analysis.
+
+        Uses subprocess isolation to ensure all memory is freed after processing,
+        preventing memory leaks from C extensions like LIEF and pillow_heif.
         """
         if not self._service_config:
             raise RuntimeError("Service not properly initialized. Call setup() first.")
 
+        process = multiprocessing.get_context("spawn").Process(
```
chromy, line 197: Saw this in passing and I'm super in favour of this! I think it will solve a lot of problems, and also be good for Android. Rather than using `multiprocessing` how do you feel about calling `subprocess.check_call("launchpad", "size", path_to_app, "-o", analysis_result)` directly and still doing the upload from the main process?

My reasoning being:
- We have all kinds of weird threads and network connections (kafka, arroyo, datadog, sentry client, sentry sdk, etc) which are the kinds of things that behave badly with fork/spawn
- If we reuse the local size command line that's one less code path to worry about
- We can keep central control over the connection to the monolith which is nice for logging / retries / queuing / kill switches etc.

trevor-e: Great suggestions.

For context Nico and I were speaking about this offline and I asked him to postpone merging this for now. I was thinking it may be easier to debug the memory issues we've been seeing while everything is still single-process, and then try to merge this in afterwards. 

I'm also not clear how this would fit in with one consumer potentially processing multiple messages in parallel, which is similar to how the iOS infra worked at Emerge. I see that Arroyo actually supports something called `RunTaskWithMultiprocessing` which might handle this automatically for us? Currently we are using `RunTaskInThreads` so I'm not sure if we've experimented with that before, it does seem like something the framework could handle for us.

NicoHinderling:
implementing this via https://github.com/getsentry/launchpad/pull/392 instead

PR #358, opened 2025-09-17
Add run configuration for memray

trevor-e:
Adding a built-in way to run `memray` against our project which is a neat memory profiling tool Nico shared.


chromy (approved):
fancy! lgtm thanks!

PR #369, opened 2025-09-19
Add extra handling for iOS asset processing

NicoHinderling:
Could be potentially helpful for this sentry error: https://sentry.sentry.io/issues/6864810165/?project=4509667577233409&query=is%3Aunresolved&referrer=issue-stream


NicoHinderling:
Potentially another symlink issue - will close now learning more about it from @trevor-e

PR #370, opened 2025-09-19
Improve retries when downloading in sentry_client

chromy:
Currently:
- In the event of a retry we ended up writing an invalid file due to
  not rewinding the file before resuming writing.
- We always start the download from scratch.
- We report the wrong file size

The endpoint does not support 'Range' but we can rewind the file so
that we end up with the correct file. At the same time we increase the number
of retries and also update some of the error handling and logging.

Fixes: EME-323


chromy:
@sentry

PR #376, opened 2025-09-23
experimenting with arroyo multiprocessing

NicoHinderling:
(proof of concept - mostly to demonstrate that to do this approach, we'd want to consider extracting the required processing logic from service into a separate class so that the kafka consumer could own the processing logic and not have to pickle the `service.process_artifact` method


NicoHinderling:
implementing this via https://github.com/getsentry/launchpad/pull/392 instead

PR #379, opened 2025-09-25
Add resumable download logic

NicoHinderling:
When testing yesterday, I uploaded larger uploads like [Firefox](https://www.emergetools.com/build/85d4934e-cc53-4efd-b6b6-27e8002eaacb) and [SF Duel](https://www.emergetools.com/build/64224eb7-53b8-4b60-992f-6d340042c747) and realized that the download logic is still breaking consistently with large enough files

This adds resumable chunk downloading using HTTP. It's working well locally

Related Sentry PR: https://github.com/getsentry/sentry/pull/100274

### DIFF
--- a/src/launchpad/sentry_client.py
+++ b/src/launchpad/sentry_client.py
@@ -212,8 +212,8 @@ class SentryClient:
                         else:
                             logger.info(f"Downloaded chunk {chunk_count} - {file_size / MB_DIVISOR:.1f} MB")
 
+                out.flush()
                 return file_size
-
             except (ConnectionError, Timeout, ChunkedEncodingError, ContentDecodingError) as e:
                 if attempt < RETRY_ATTEMPTS - 1:
                     wait_time = 5


src/launchpad/sentry_client.py
```diff
@@ -186,6 +255,122 @@ def download_artifact_to_file(self, org: str, project: str, artifact_id: str, ou
...
+                        logger.error(f"Failed to download chunk after 3 attempts: {e}")
+                        # Don't raise immediately, let's try the fallback
+                        break
+
+            if not chunk_success:
+                logger.error(f"Failed to download chunk at byte {start}, falling back to standard download")
+                # Reset file position and try standard download from beginning
+                out.seek(0)
+                out.truncate()
+                return self.download_artifact_to_file(org, project, artifact_id, out)
```
chromy, line 369: I don't think it is worth supporting both. It gets quite complicated it seems fine to mandate the endpoint supports Range (once you land the Sentry side PR). We could even support both in simpler way:
- Try and download everything
- If that fails and endpoint supports Range -> use Range to resume the download
- If that fails and the endpoint does not not support Range -> raise

That way we still support non-range - we just lose the retrying and if you want the retrying you should support Range.
   - use range qury

src/launchpad/sentry_client.py
```diff
@@ -186,6 +255,122 @@ def download_artifact_to_file(self, org: str, project: str, artifact_id: str, ou
 
         return file_size
 
+    def download_artifact_resumable(self, org: str, project: str, artifact_id: str, out: io.BytesIO) -> int:
```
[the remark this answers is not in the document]
NicoHinderling: yes you're spot on. I've simplified it considerably since then. Hopefully it looks better now 🤞

src/launchpad/sentry_client.py
```diff
@@ -138,53 +142,62 @@ def __init__(self, base_url: str, shared_secret: str | None = None) -> None:
...
-        return file_size
+            except (ConnectionError, Timeout, ChunkedEncodingError, ContentDecodingError) as e:
+                if attempt < RETRY_ATTEMPTS - 1:
+                    wait_time = 2**attempt  # Exponential backoff
+                    logger.warning(
+                        f"Download failed (attempt {attempt + 1}/{RETRY_ATTEMPTS}), retrying in {wait_time}s: {e}"
+                    )
+                    time.sleep(wait_time)
+                else:
+                    raise SentryClientError(exception=e, detail="Download failed after retries")
```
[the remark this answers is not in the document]
NicoHinderling: not going to worry about the case where the endpoint doesn't support ranges since we control the endpoint

src/launchpad/sentry_client.py
```diff
@@ -138,53 +142,65 @@ def __init__(self, base_url: str, shared_secret: str | None = None) -> None:
...
-                response = self.session.get(url, auth=self.auth, timeout=120, stream=True)
-                if response.status_code != 200:
+                headers = {}
+                if file_size > 0:
+                    headers = {"Range": f"bytes={file_size}-"}
+
+                response = self.session.get(url, auth=self.auth, headers=headers, timeout=120, stream=True)
+
+                if response.status_code not in (200, 206):
                     raise SentryClientError(response=response)
```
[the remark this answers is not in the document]
NicoHinderling: Assumes that the server doesnt support range requests , which it always should and will

src/launchpad/sentry_client.py
```diff
@@ -138,53 +142,65 @@ def __init__(self, base_url: str, shared_secret: str | None = None) -> None:
...
-                else:
-                    logger.exception(f"Download failed after {attempts} attempts due to network error")
-                    raise SentryClientError(
-                        detail="Failed to download artifact (network error)", response=response, exception=e
+                if attempt < RETRY_ATTEMPTS - 1:
+                    wait_time = 2**attempt  # Exponential backoff
+                    logger.warning(
+                        f"Download failed (attempt {attempt + 1}/{RETRY_ATTEMPTS}), retrying in {wait_time}s: {e}"
                     )
+                    time.sleep(wait_time)
```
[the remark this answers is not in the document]
NicoHinderling: Assumes that the server doesnt support resumable downloads, which it always should and will

src/launchpad/sentry_client.py
```diff
@@ -138,53 +142,65 @@ def __init__(self, base_url: str, shared_secret: str | None = None) -> None:
...
+
             except (ConnectionError, Timeout, ChunkedEncodingError, ContentDecodingError) as e:
-                if attempt == 0:
-                    logger.warning(f"Download failed due to network error ({attempt + 1}/{attempts})", exc_info=True)
-                else:
-                    logger.exception(f"Download failed after {attempts} attempts due to network error")
-                    raise SentryClientError(
-                        detail="Failed to download artifact (network error)", response=response, exception=e
+                if attempt < RETRY_ATTEMPTS - 1:
+                    wait_time = 2**attempt  # Exponential backoff
```
trevor-e, line 194: N is so small this won't really provide exponential backoff like you'd expect. It might be better to just use a fixed wait time like 5 seconds.

tests/unit/test_sentry_client.py
```diff
@@ -283,3 +284,170 @@ def test_upload_installable_app_single_chunk_failure(self):
...
+            responses.HEAD,
+            "https://example.com/api/0/internal/test-org/test-project/files/preprodartifacts/test-artifact/",
+            headers={"Content-Length": "40"},  # 40 bytes total
+        )
+
+        # Mock GET request for download
+        responses.add(
+            responses.GET,
+            "https://example.com/api/0/internal/test-org/test-project/files/preprodartifacts/test-artifact/",
+            body=b"A" * 20 + b"B" * 20,  # 40 bytes total, will be 2 chunks of 20MB each (but truncated)
```
trevor-e, line 302: how does this become two chunks?

tests/unit/test_sentry_client.py
```diff
@@ -283,3 +284,170 @@ def test_upload_installable_app_single_chunk_failure(self):
...
+
+        result = client.download_artifact("test-org", "test-project", "test-artifact", out)
+
+        assert result == 40
+        out.seek(0)
+        assert out.read() == b"A" * 20 + b"B" * 20
+
+    @responses.activate
+    def test_download_artifact_no_head_request(self):
+        """Test download when HEAD request fails."""
```
trevor-e, line 316: In what case could the HEAD request fail but the GET request succeed? If the HEAD request fails couldn't we just mark the entire thing as failed?

tests/unit/test_sentry_client.py
```diff
@@ -283,3 +284,170 @@ def test_upload_installable_app_single_chunk_failure(self):
...
+        """Test download fails after maximum retries."""
+        # Mock HEAD request for file size
+        responses.add(
+            responses.HEAD,
+            "https://example.com/api/0/internal/test-org/test-project/files/preprodartifacts/test-artifact/",
+            headers={"Content-Length": "13"},
+        )
+
+        # All attempts fail with ConnectionError
+        for _ in range(3):  # RETRY_ATTEMPTS = 3
```
trevor-e, line 440: Can this use `RETRY_ATTEMPTS` in case we end up changing that?

src/launchpad/sentry_client.py
```diff
@@ -138,53 +142,65 @@ def __init__(self, base_url: str, shared_secret: str | None = None) -> None:
...
+            head_resp = self.session.head(url, auth=self.auth, timeout=30)
+            if head_resp.status_code == 200:
+                content_length = head_resp.headers.get("Content-Length")
+                if content_length:
+                    total_size = int(content_length)
+        except Exception:
+            pass
+
         file_size = 0
+        chunk_size = 20 * 1024 * 1024  # 20MB chunks
```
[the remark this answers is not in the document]
NicoHinderling: I'm open to trying a smaller chunking size but I think this is worth a try!

tests/unit/test_sentry_client.py
NicoHinderling: sorry, the comment is not correct. updated/removed

src/launchpad/sentry_client.py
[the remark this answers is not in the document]
trevor-e: Makes sense to me, might be worth including that docs link here? If we fine-tune these values more in the future it's helpful to know how they were chosen.

tests/unit/test_sentry_client.py
NicoHinderling: the HEAD request is _technically_ optional, as it's only used to get the Content-Length header for progress reporting

Frankly the only valid reason where HEAD fails but GET succeeds would be due to some transient network issue. Otherwise I expect them to both succeed, it just seemed better to allow HEAD to be optional

trevor-e: It's technically optional in a vacuum, but since we are the ones controlling the business logic can't we make it required? 😆 I'm just a bit confused why we are supporting so many different configurations unless there's something I'm missing.

trevor-e (approved):

PR #380, opened 2025-09-25
ref(preprod): clean up duplicate exception logging

trevor-e:
I'm seeing duplicate issues get created, e.g. https://sentry.sentry.io/issues/6901677123/?project=4509667577233409&query=is%3Aunresolved&referrer=issue-stream and https://sentry.sentry.io/issues/6901677118/?project=4509667577233409&query=is%3Aunresolved&referrer=issue-stream

I also went through and got rid of some unnecessary `{e}` printing since the `exc_info` automatically captures that.

### DIFF
--- a/src/launchpad/artifacts/apple/zipped_xcarchive.py
+++ b/src/launchpad/artifacts/apple/zipped_xcarchive.py
@@ -66,7 +66,7 @@ class ZippedXCArchive(AppleArtifact):
             self._plist = plist_data
             return plist_data
         except Exception as e:
-            raise RuntimeError(f"Failed to parse Info.plist: {e}")
+            raise RuntimeError("Failed to parse Info.plist") from e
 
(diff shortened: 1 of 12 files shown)


src/launchpad/artifacts/apple/zipped_xcarchive.py
```diff
@@ -295,8 +295,8 @@ def get_asset_catalog_details(self, relative_path: Path) -> List[AssetCatalogEle
                 data = json.load(f)
 
             return [self._parse_asset_element(item, parent_path) for item in data]
-        except Exception as e:
-            logger.error(f"Failed to get asset catalog details for {relative_path}: {e}")
+        except Exception:
+            logger.exception(f"Failed to get asset catalog details for {relative_path}")
```
trevor-e, line 299: I'll remove the paths in here once we get more users, right now seems worth it to create new issues.

PR #389, opened 2025-09-30
Add check_reasonable_zip

chromy:
One of the main things we do is parse ZIPs and evaluate their contents.
This (and more generally parsing untrusted input) is the major attack
surface for launchpad. There are a couple well known avenues for
malicious zips:
1. ZIP bombs, where a small ZIP expands to many TB/PB when unzipped
   using fancy aliasing tricks.
2. Abusing absolute and relative paths to overwrite files the unzipper
   did not intend.
3. Being very large / having many files when expanded.

- 1 and 2 are mitigated by Python
- 2 was is also prevented by additional code in ZipProvider
- 3 is resolved by this PR which adds limits on the number of files
   (100,000) and maximum unzipped size (10GB).

These limits have to be quite large (since people will mostly want to
use launchpad on large apps) but should be small enough to avoid people
taking down the whole consumer by uploading a carefully constructed
file.

See also: https://www.bamsoftware.com/hacks/zipbomb/

Resolves: EME-381

### DIFF
--- a/src/launchpad/artifacts/providers/zip_provider.py
+++ b/src/launchpad/artifacts/providers/zip_provider.py
@@ -8,6 +8,65 @@ from launchpad.utils.logging import get_logger
 
 logger = get_logger(__name__)
 
+DEFAULT_MAX_FILE_COUNT = 100000
+DEFAULT_MAX_UNCOMPRESSED_SIZE = 10 * 1024 * 1024 * 1024
+
+
+class UnreasonableZipError(ValueError):
+    """Raised when a zip file exceeds reasonable limits."""
+
+    pass
+
+
+class UnsafePathError(ValueError):
+    """Raised when a zip file contains unsafe path entries that could lead to path traversal attacks."""
+
+    pass
+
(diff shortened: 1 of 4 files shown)


src/launchpad/artifacts/providers/zip_provider.py
```diff
@@ -8,6 +8,59 @@
 
 logger = get_logger(__name__)
 
+DEFAULT_MAX_FILE_COUNT = 100000
+DEFAULT_MAX_UNCOMPRESSED_SIZE = 10 * 1024 * 1024 * 1024
+
+
+class UnreasonableZipError(ValueError):
```
trevor-e, line 15: nit: Maybe the traversal logic should have its own error subclass as well?

trevor-e (approved):

src/launchpad/artifacts/providers/zip_provider.py
chromy: done, thanks!

PR #392, opened 2025-09-30
Configure arroyo multiprocessing setup

NicoHinderling:
https://github.com/getsentry/launchpad/pull/354 has a better description of the motivation behind exploring multiprocessing. This is a different attempt at it but instead using the arroyo multiprocessing consumer setup to accomplish it

### DIFF
--- a/src/launchpad/kafka.py
+++ b/src/launchpad/kafka.py
@@ -3,7 +3,9 @@
 from __future__ import annotations
 
 import asyncio
+import logging
 import os
+import signal
 import time
 
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


src/launchpad/kafka.py
```diff
@@ -16,23 +18,52 @@
...
 # Schema codec for preprod artifact events
 PREPROD_ARTIFACT_SCHEMA = get_codec(PREPROD_ARTIFACT_EVENTS_TOPIC)
 
 
+def process_kafka_message_with_service(msg: Message[KafkaPayload]) -> Any:
+    """Process a Kafka message using the actual service logic in a worker process."""
+
+    def worker_signal_handler(signum: int, frame) -> None:
+        logger.info(f"Worker process {os.getpid()} received signal {signum}, exiting...")
+        os._exit(0)
```
chromy, line 49: We probably want to exit with a non-zero code to signal the failure

src/launchpad/kafka.py
```diff
@@ -16,23 +18,52 @@
...
+    if not logging.getLogger().handlers:
+        server_config = get_server_config()
+        setup_logging(verbose=server_config.debug, quiet=not server_config.debug)
+
+    try:
+        decoded = PREPROD_ARTIFACT_SCHEMA.decode(msg.payload.value)
+        ArtifactProcessor.process_message(decoded)
+        return decoded  # type: ignore[no-any-return]
+    except Exception as e:
+        logger.error(f"Failed to process message in worker: {e}", exc_info=True)
```
chromy, line 63: I learned recently that `logger.error(..., exc_info=True)` is the same as `logger.exception(...)` so we should use the latter in these situations.

src/launchpad/kafka.py
```diff
@@ -228,7 +292,7 @@ def get_kafka_config() -> KafkaConfig:
         bootstrap_servers=bootstrap_servers,
         group_id=group_id,
         topics=topics_env.split(","),
-        concurrency=int(os.getenv("KAFKA_CONCURRENCY", "4")),
+        concurrency=int(os.getenv("KAFKA_CONCURRENCY", "2")),
         max_pending_futures=int(os.getenv("KAFKA_MAX_PENDING_FUTURES", "100")),
```
chromy, line 296: I think max_pending_futures is unused now?

src/launchpad/kafka.py
```diff
@@ -122,16 +152,57 @@ def run(self):
         finally:
             self._running = False
```
chromy, line 153: I think `run()` is unused now

src/launchpad/kafka.py
NicoHinderling: true

PR #399, opened 2025-10-01
Try to force subprocess logs to write to sdout

NicoHinderling:
Another potential fix for subprocess logs to show up in GCP

In GCP/Docker, logs are captured from the container's main process (PID 1), and spawned subprocesses might not have their stdout/stderr properly connected, despite our previous attempts

If this doesn't work, then I will implement a logging queue

### DIFF
--- a/src/launchpad/kafka.py
+++ b/src/launchpad/kafka.py
@@ -2,9 +2,11 @@
 
 from __future__ import annotations
 
+import logging
 import multiprocessing
 import os
 import signal
+import sys
 import time
 
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


src/launchpad/kafka.py
```diff
@@ -205,7 +205,20 @@ class LaunchpadStrategyFactory(ProcessingStrategyFactory[KafkaPayload]):
...
     @staticmethod
     def _initialize_worker_logging() -> None:
-        """Initialize logging in worker process."""
+        """Initialize logging in worker process.
+
+        With multiprocessing spawn context, subprocesses don't inherit
+        parent's stdout/stderr. We need to explicitly configure logging
+        to write to stdout for Docker/GCP log collection.
+        """
+        import sys
```
rbro112, line 214: Any reason to not add this top-level (genuine question too, my python knowledge ain't deep enough to know)

NicoHinderling: no, it should be at the top. i'm just lazy and on my 3rd attempt. I have little faith in this but want to try it before doing a more complex approach lol - but ya I'll move it , all g

PR #401, opened 2025-10-02
Try logging queue for subprocess logs in GCP

NicoHinderling:
With multiprocessing spawn context, worker subprocesses have their own stdout/stderr that Docker/GCP doesn't capture. Only the main process's output streams are monitored, despite my previous attempts (workers logged directly to their disconnected stdout, etc).

With this, workers send log records through a queue to the main process, which formats and writes them to its stdout that Docker is actually capturing

### DIFF
--- a/src/launchpad/kafka.py
+++ b/src/launchpad/kafka.py
@@ -6,11 +6,11 @@ import logging
 import multiprocessing
 import os
 import signal
-import sys
 import time
 
 from dataclasses import dataclass
 from functools import partial
+from logging.handlers import QueueHandler, QueueListener
 from multiprocessing.pool import Pool
 from typing import Any, Callable, Mapping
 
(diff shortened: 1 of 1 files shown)


src/launchpad/kafka.py
```diff
@@ -205,39 +204,48 @@ def is_healthy(self) -> bool:
...
+    @staticmethod
+    def _initialize_worker_logging(log_queue: multiprocessing.Queue[Any]) -> None:
+        """Initialize logging in worker process to send logs to queue.
+
+        With multiprocessing spawn context, subprocesses don't inherit
+        parent's stdout/stderr. We use a queue to send log records to
+        the main process which writes them to stdout for Docker/GCP.
+        """
+        root_logger = logging.getLogger()
+        root_logger.handlers.clear()
```
trevor-e, line 243: Is this going to cause race conditions with one worker process wiping out the logger for another worker process? Not super clear to me why this has to be cleared.

trevor-e (approved):

PR #417, opened 2025-10-15
Add "optimize alternative icons" insight handling

NicoHinderling:
Closes EME-50

I tried to match the emerge implementation as much as possible. There's a ~6% size difference between the ultimately optimized icons compared to what's produced by the emerge code But it is consistent across all the images and appears to be related to the fact that we are using a different HEIC encoder:

- in emerge we use Apple's native HEIC encoding via swift (Core Graphics/ImageIO)
- in launchpad we use libheif

I tried reducing the image quality from 85% to 75% to get the size reduction closer to emerge's but it didn't reduce the size enough to justify an extra 10% loss in quality

### DIFF
--- a/src/launchpad/size/insights/apple/image_optimization.py
+++ b/src/launchpad/size/insights/apple/image_optimization.py
@@ -79,9 +79,8 @@ class BaseImageOptimizationInsight(Insight[ImageOptimizationInsightResult], ABC)
                     result = future.result()
                     if result and result.potential_savings >= self.MIN_SAVINGS_THRESHOLD:
                         results.append(result)
-                except Exception as exc:  # pragma: no cover
-                    file_info = future_to_file[future]
-                    logger.error("Failed to analyse %s: %s", file_info.path, exc)
+                except Exception:  # pragma: no cover
+                    logger.exception("Failed to analyze image in thread pool")
 
         if not results:
             return None
@@ -138,8 +137,8 @@ class BaseImageOptimizationInsight(Insight[ImageOptimizationInsightResult], ABC)
                     conversion_savings=total_conversion,
                     heic_size=heic_size,
                 )
-        except Exception as exc:
-            logger.error("Failed to process %s: %s", file_info.path, exc)
+        except Exception:
+            logger.exception("Failed to open or process image file")
             return None
 
     def _check_minification(self, img: Image.Image, file_size: int, fmt: str) -> _OptimizationResult | None:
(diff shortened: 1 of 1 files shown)


src/launchpad/size/insights/apple/alternate_icons_optimization.py
```diff
@@ -0,0 +1,133 @@
...
+
+    Alternate app icons can be optimized without affecting the App Store listing since
+    only the primary icon is displayed there. This insight identifies alternate icons
+    that could be minified or converted to more efficient formats like HEIC.
+
+    Icons are resized to device display size (180px for iPhone 3x) and back to store
+    size (1024px) before optimization, since they only need quality for homescreen display.
+    """
+
+    # Icon size constants matching Swift implementation
```
trevor-e: ai generated comment? lol

src/launchpad/size/insights/apple/alternate_icons_optimization.py
```diff
@@ -0,0 +1,133 @@
...
+            if fmt in {"png", "jpg", "jpeg"}:
+                if res := self._check_minification(img, resized_size, fmt):
+                    minify_savings, minified_size = res.savings, res.optimized_size
+                if res := self._check_heic_conversion(img, resized_size):
+                    conversion_savings, heic_size = res.savings, res.optimized_size
+            elif fmt in {"heif", "heic"}:
+                if res := self._check_heic_minification(img, resized_size):
+                    minify_savings, minified_size = res.savings, res.optimized_size
+        except Exception as exc:
+            logger.error("Failed to analyze resized icon: %s", exc)
```
trevor-e: Same, `logger.exception()`

src/launchpad/size/insights/apple/image_optimization.py
```diff
@@ -36,17 +37,26 @@ class _OptimizationResult:
...
+    """Base class for image optimization insights with shared analysis logic."""
 
     OPTIMIZABLE_FORMATS = {"png", "jpg", "jpeg", "heif", "heic"}
     MIN_SAVINGS_THRESHOLD = 4096
     TARGET_JPEG_QUALITY = 85
     TARGET_HEIC_QUALITY = 85
     _MAX_WORKERS = 4
 
+    @abstractmethod
+    def _iter_files_to_analyze(self, input: InsightsInput) -> Iterable[FileInfo]:
```
trevor-e: Can these two methods be combined into something like `_find_images() -> List[FileInfo]`? The de-duping can be merged into that as well. The current structure here is a bit awkward IMO, the iterator just converts it to a list anyways. It looks like the subclasses are responsible for returning a list of images that should be considered for processing.

trevor-e: Similarly, the `_analyze_image_optimization` override doesn't fit in nicely here since we lose the ThreadPoolExecutor logic and duplicate a lot of size-checking code. Maybe there could be some kind of overridable "preprocess" or "transform" method that gets called before `_analyze_image_optimization` so that method doesn't have to be overriden?

PR #420, opened 2025-10-20
chore(preprod): Save images as original file format to ParsedAssets

NicoHinderling:
If https://github.com/getsentry/sentry-cli/pull/2859 lands, this logic can be as we originally expected it: looking for a parsed asset with the original file ending


runningcode (approved):
LGTM, Should we add a test for this?
Are we assuming that no clients will be using older versions of the CLI? I guess that is reasonable since we are still before EA.

PR #425, opened 2025-10-22
Improve symbol categorization and aggregation logic

trevor-e:
This is a pretty massive overhaul of how our symbol size calculation + treemap generation works for iOS, it should be more accurate and extensible. I find it a lot easier to follow now too. This also **adds a couple nodes** to our treemap, notably a new "Other Symbols" node since I saw that many C/C++ symbols were taking up quite a bit of size.

There were a few issues with the old code, namely that there was potential for a symbol to be counted twice since each aggregator was looping over the entire list. The process works like this:
1. We calculate the symbol sizes like normal
2. In a single loop, we categorize and partition each symbol to tell if it's Swift/ObjC/C++/Compiler Generated/Other
3. For each partition, we have aggregators which combine the symbols into groups, most often the compile unit they belong to e.g. a Swift type.

Afterwards, we take these aggregated groups and build a treemap out of them. The treemap code works like this:
1. We take each of the segment/sections for the Mach-O binary and their original sizes
2. As we loop over each symbol group to add to the treemap, we debit size from the appropriate binary section at the same time so that our size accounting nets out
3. After adding all of our types to the treemap, we add any segments/sections that still have leftover size we weren't able to attribute

Screenshot 2025-10-23 at 5 22 08 PM

### DIFF
--- a/src/launchpad/size/analyzers/apple.py
+++ b/src/launchpad/size/analyzers/apple.py
@@ -460,12 +460,6 @@ class AppleAppAnalyzer:
                     logger.debug(
                         f"Parsed {len(dwarf_relocations.relocations)} DWARF relocations for {binary_info.name}"
                     )
-                    # Validate symbol sizes if we have both symbol_sizes and relocations
-                    # if symbol_sizes:
-                    #     with sentry_sdk.start_span(op="validate", description="symbol_sizes_vs_relocations"):
-                    #         self._validate_symbol_sizes_against_relocations(
-                    #             binary_info.name, symbol_sizes, dwarf_relocations
-                    #         )
         else:
             logger.info(
                 "size.apple.skip_symbol_analysis.no_dwarf_binary",
--- a/src/launchpad/size/symbols/cpp_aggregator.py
+++ b/src/launchpad/size/symbols/cpp_aggregator.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import re
 
 from collections import defaultdict
-from typing import List, NamedTuple, Tuple
+from typing import List, NamedTuple
 
 import sentry_sdk
 
@@ -18,7 +18,7 @@ logger = get_logger(__name__)
 
 class CppNamespaceFunction(NamedTuple):
     namespace: str
-    function_name: str | None  # None for namespace-level symbols
+    function_name: str
 
 
 class CppSymbolTypeAggregator:
@@ -43,18 +43,12 @@ class CppSymbolTypeAggregator:
         return mangled_name.startswith("_Z") or mangled_name.startswith("__Z")
 
     @staticmethod
-    def _extract_namespace_and_function(mangled_name: str) -> Tuple[str, str]:
-        """
-        Extract simplified namespace and function name from C++ mangled name.
-
-        Returns:
-            Tuple of (namespace, function_name)
-        """
+    def _extract_namespace_and_function(mangled_name: str) -> CppNamespaceFunction:
         # Try to extract nested name components
         # Format: __ZN<length><name><length><name>...<length><function>E...
         if not (mangled_name.startswith("_ZN") or mangled_name.startswith("__ZN")):
             # Not a nested name, treat as global namespace
-            return ("(global)", mangled_name)
+            return CppNamespaceFunction(namespace="(global)", function_name=mangled_name)
 
         # Skip the __ZN or _ZN prefix
         rest = mangled_name[4:] if mangled_name.startswith("__ZN") else mangled_name[3:]
@@ -86,11 +80,11 @@ class CppSymbolTypeAggregator:
                 break
 
         if not components:
-            return ("(unknown)", mangled_name)
+            return CppNamespaceFunction(namespace="(unknown)", function_name=mangled_name)
 
         # Last component is usually the function name
         if len(components) == 1:
-            return ("(global)", components[0])
+            return CppNamespaceFunction(namespace="(global)", function_name=components[0])
 
         # Build namespace from all but last component
         namespace = "::".join(components[:-1])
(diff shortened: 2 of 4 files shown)


src/launchpad/size/analyzers/apple.py
```diff
@@ -505,6 +495,8 @@ def _analyze_binary(
             header_size=parser.get_header_size(),
             dyld_info=dyld_info,
             dwarf_relocations=dwarf_relocations,
+            static_inits=static_inits,
```
trevor-e: I moved these out of `SymbolInfo` since they weren't really related.

src/launchpad/size/treemap/macho_element_builder.py
```diff
@@ -59,220 +57,383 @@ def build_element(self, file_info: FileInfo, display_name: str) -> TreemapElemen
...
+            key = canonical_key(seg_name, sec_name)
+            if not key or key not in section_remaining:
+                return 0
+            take = min(sz, section_remaining[key])
+            if take:
+                section_remaining[key] -= take
+            return take
+
         section_subtractions: Dict[str, int] = {}
-        segment_subtractions: Dict[str, int] = {}
```
trevor-e: These changes look a lot more involved than they actually are. Mostly I changed over to the "debit" terminology since I kept getting confused with the subtractions. And then I refactored this massive function into smaller functions since I wanted to add the "Other Symbols" node.

src/launchpad/size/symbols/cpp_aggregator.py
```diff
@@ -0,0 +1,130 @@
...
+    )
+
+    @staticmethod
+    def is_cpp_symbol(mangled_name: str) -> bool:
+        """Check if a symbol is a C++ symbol based on its mangled name."""
+        # C++ symbols often use Itanium mangling: _Z or __Z prefix
+        return mangled_name.startswith("_Z") or mangled_name.startswith("__Z")
+
+    @staticmethod
+    def _extract_namespace_and_function(mangled_name: str) -> Tuple[str, str]:
```
NicoHinderling: nit: why doesnt this method just return a `CppNamespaceFunction` type?

src/launchpad/size/symbols/swift_aggregator.py
```diff
@@ -0,0 +1,94 @@
...
+        Includes:
+        - _$s: Modern Swift mangling
+        - _Tt: Older Swift mangling
+        - __IVARS__Tt: ObjC runtime metadata for Swift classes
+        - __DATA__Tt: Data section symbols for Swift types
+        """
+        if mangled_name.startswith("_$s") or mangled_name.startswith("_Tt"):
+            return True
+        # Swift classes exposed to ObjC have metadata with Swift mangling
+        if "__Tt" in mangled_name or "__TtC" in mangled_name:
```
NicoHinderling: isn't `"__TtC" in mangled_name` redundant since `"__Tt" in mangled_name` would always be true in that case too?

src/launchpad/size/symbols/swift_aggregator.py
```diff
@@ -0,0 +1,94 @@
...
+                type_name = "Unattributed"
+
+            key = SwiftModuleType(module=module, type_name=type_name)
+            if key not in type_groups:
+                type_groups[key] = []
+            type_groups[key].append(symbol)
+
+        result: list[SwiftSymbolTypeGroup] = []
+        for key, symbols in type_groups.items():
+            symbols.sort(key=lambda x: x.size, reverse=True)
```
NicoHinderling: just curious, how come these need to be sorted? i noticed the symbols for cpp and objc aren't sorted

src/launchpad/size/treemap/macho_element_builder.py
```diff
@@ -59,220 +57,372 @@ def build_element(self, file_info: FileInfo, display_name: str) -> TreemapElemen
...
-                    if adjusted <= 0:
-                        logger.debug(
-                            f"Skipping section {unique_sec} - no remaining size {adjusted} after symbol subtraction"
+                    key = f"{segment_name}.{section_name}"
+                    subtraction = section_subtractions.get(key, 0)
+                    if subtraction > section_size:
+                        logger.warning(
+                            f"Section {key}: symbol bytes ({subtraction:,}) exceed section size ({section_size:,})."
                         )
+                        subtraction = section_size
```
NicoHinderling: would a miscalculation situation like this ever warrant a sentry error for us? 

i feel like this sort of thing is more significant than a log warning , no?

src/launchpad/size/treemap/macho_element_builder.py
```diff
@@ -337,24 +479,21 @@ def _build_metadata_components(self, binary_analysis: MachOBinaryAnalysis) -> Li
...
+                    name=lc.name,
+                    size=lc.size,
+                    type=TreemapType.EXECUTABLES,
+                    path=None,
+                    is_dir=False,
+                    children=[],
                 )
-
-            total_lc_size = sum(lc.size for lc in binary_analysis.load_commands)
+                for lc in binary_analysis.load_commands
```
NicoHinderling: mr.pythonista

src/launchpad/size/symbols/swift_aggregator.py
trevor-e: Was useful during debugging so just left it in, can do the same for the others.

trevor-e: Actually will just remove for now.

src/launchpad/size/symbols/cpp_aggregator.py
```diff
@@ -0,0 +1,130 @@
...
+from launchpad.size.symbols.macho_symbol_sizes import SymbolSize
+from launchpad.size.symbols.types import CppSymbolList, CppSymbolTypeGroup
+from launchpad.utils.logging import get_logger
+
+logger = get_logger(__name__)
+
+
+class CppNamespaceFunction(NamedTuple):
+    namespace: str
+    function_name: str | None  # None for namespace-level symbols
```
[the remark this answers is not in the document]
trevor-e: fixed to not be optional

PR #429, opened 2025-10-24
Publish docker image to multi-region image registry too

NicoHinderling:
Closes EME-464

I think we'll later need to update the following two files if the single region image registry gets removed:

- [deploy.sh](https://github.com/getsentry/launchpad/blob/main/gocd/templates/bash/deploy.sh)
- [check-cloudbuild.sh](https://github.com/getsentry/launchpad/blob/main/gocd/templates/bash/check-cloudbuild.sh)

since I'm pretty sure they all rely on the single region image atm

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -26,19 +26,19 @@ steps:
             --push \
             --platform linux/amd64 \
             --build-arg LAUNCHPAD_VERSION_SHA=$COMMIT_SHA \
-            -t $LOCATION-docker.pkg.dev/$PROJECT_ID/$REPO_NAME/image:$COMMIT_SHA \
-            -t $LOCATION-docker.pkg.dev/$PROJECT_ID/$REPO_NAME/image:$SHORT_SHA \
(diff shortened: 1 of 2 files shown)


cloudbuild.yaml
```diff
@@ -40,4 +40,31 @@ steps:
...
+            --label org.opencontainers.image.revision=$COMMIT_SHA \
+            --label org.opencontainers.image.version=$COMMIT_SHA \
+            --label org.opencontainers.image.title=$REPO_NAME \
+            --label org.opencontainers.image.source="https://github.com/$REPO_FULL_NAME" \
+            --label org.opencontainers.vendor="Sentry" \
+            --label org.opencontainers.image.url=$REPO_FULL_NAME \
+            --cache-from $LOCATION-docker.pkg.dev/$PROJECT_ID/$REPO_NAME-mr/image:latest \
+            --sbom true \
+            --provenance="mode=max" \
+            .
```
mwarkentin: Up to you, but you don't have to duplicate this whole thing, you can add extra tags with `-t` to the existing `docker buildx build` job.

PR #437, opened 2025-10-28
Ensure all asset catalog images are considered

NicoHinderling:
When debugging why alternative icons are not showing up for Firefox, I realized that some asset catalog images will not have a file path because they are multi-sized assets. I've made a change in the CLI to include them into ParsedAssets still, but we need to ensure that they are parsed with the correct path on launchpad to ensure they are considered and fully analyzed for our image insights

For example, here's one in the firefox app that we weren't actually considering:

```
    {
        "vector": false,
        "type": 0,
        "imageId": "42D31467-E63B-49D8-8D9B-1B608059E344",
        "size": 236,
        "height": 1024,
        "name": "AppIcon_Alt_Color_Purple",
        "filename": "AppIcon_Alt_Color_Purple",
        "idiom": "phone",
        "width": 1024
    },
```

### DIFF
--- a/src/launchpad/size/insights/apple/alternate_icons_optimization.py
+++ b/src/launchpad/size/insights/apple/alternate_icons_optimization.py
@@ -71,7 +71,6 @@ class AlternateIconsOptimizationInsight(BaseImageOptimizationInsight):
         )
 
     def _is_alternate_icon_file(self, file_info: FileInfo, alternate_icon_names: set[str]) -> bool:
-        return (
-            file_info.file_type.lower() in self.OPTIMIZABLE_FORMATS
-            and Path(file_info.path).stem in alternate_icon_names
-        )
+        # Some asset catalog entries have no extension, so we include "other" in the OPTIMIZABLE_FORMATS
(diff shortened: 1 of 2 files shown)


src/launchpad/size/insights/apple/alternate_icons_optimization.py
```diff
@@ -71,7 +71,6 @@ def _resize_icon_for_analysis(self, img: Image.Image) -> Image.Image:
         )
 
     def _is_alternate_icon_file(self, file_info: FileInfo, alternate_icon_names: set[str]) -> bool:
-        return (
-            file_info.file_type.lower() in self.OPTIMIZABLE_FORMATS
-            and Path(file_info.path).stem in alternate_icon_names
-        )
+        # Some asset catalog entries have no extension, so we include "other" in the OPTIMIZABLE_FORMATS
+        file_type_match = file_info.file_type.lower() in (self.OPTIMIZABLE_FORMATS | {"other"})
```
trevor-e: The "other" check doesn't seem ideal, is file_type not showing as PNG for these?

PR #447, opened 2025-10-30
Fix minify strings insight

trevor-e:
There were a couple issues around plist handling:
- Binary plists were being incorrectly flagged, these can be safely skipped since the format doesn't allow for comments
- We also weren't handling regular plists properly. Now we load the plist first before applying the minification


src/launchpad/size/insights/apple/localized_strings_minify.py
```diff
@@ -12,51 +11,24 @@
 logger = logging.getLogger(__name__)
 
 
-class MinifyLocalizedStringsProcessor:
```
trevor-e: Moved below

PR #456, opened 2025-11-04
One more tweak to the alt icon calculation logic

NicoHinderling:
While working on the firefox PR changes and the blog post, I came across three important details:

1. If you use a HEIC for an app icon, two copies of the HEIC are included in the binary (even if thinned). This needs to be taken into account for the size calculation.
2. App icon's as PNGs don't necessary get saved the same as how they are included. I tried using the current best png compression tool `zopflipng` to compress all of the firefox alt icons and after archiving/thinning/uploading, the install size somehow slightly went up. Clearly we can't reliably just suggest to users that **specifically for app icons** they should just optimize their pngs. As such, our guidance should probably just be kept simple and focus on converting them to HEICs.
3. The PIL (python library) resizing and file conversion logic yields significantly different file sizes compared to Mac OS's `sips` utility method as well as the native Swift utils that we previously used at emerge.

This PR is mostly focused on point #3 but the first two points are extremely relevant. I basically spent a lot of last week and half of my day yesterday meticulously testing and found a decent multiplier to get our size reduction estimates closer to reality (currently quite off, 8.5mb when in reality it's 3mb for Firefox)

### DIFF
--- a/tests/unit/test_alternate_icons_optimization.py
+++ b/tests/unit/test_alternate_icons_optimization.py
@@ -99,9 +99,10 @@ class TestAlternateIconsOptimizationInsight:
     def test_identifies_alternate_icons(self):
         """Test that alternate icons are correctly identified and analyzed."""
         # Create test images (larger and unoptimized to ensure savings > 4KB)
-        primary_icon_path = self._create_test_png((200, 200), optimized=False)
-        alt_icon1_path = self._create_test_png((200, 200), optimized=False)
-        alt_icon2_path = self._create_test_png((200, 200), optimized=False)
+        # Using 1024x1024 to match actual app icon sizes
+        primary_icon_path = self._create_test_png((1024, 1024), optimized=False)
+        alt_icon1_path = self._create_test_png((1024, 1024), optimized=False)
+        alt_icon2_path = self._create_test_png((1024, 1024), optimized=False)
 
         try:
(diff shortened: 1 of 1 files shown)


src/launchpad/size/insights/apple/alternate_icons_optimization.py
```diff
@@ -53,24 +63,31 @@ def _find_images(self, input: InsightsInput) -> List[FileInfo]:
...
-
-        baseline_savings = max(0, file_info.size - resized_size)
-        return resized, resized_size, baseline_savings
-
     def _resize_icon_for_analysis(self, img: Image.Image) -> Image.Image:
         return img.resize((self.IPHONE_3X_ICON_SIZE, self.IPHONE_3X_ICON_SIZE), Image.Resampling.LANCZOS).resize(
             (self.APP_STORE_ICON_SIZE, self.APP_STORE_ICON_SIZE), Image.Resampling.LANCZOS
         )
+
+    def _check_heic_conversion(
```
NicoHinderling: the script that we show users literally does the resizing like this, so lets directly emulate that approach since it also yields the most similar outcome

NicoHinderling:
discussed offline with @trevor-e. Going to close this and put another PR up to just disable this insight altogether for now

we're deeming this insufficient since the best that I could come up with relied on a magic number (the 1.5x multiplier)

PR #464, opened 2025-11-10
feat: Extract and send tooling versions to Sentry (EME-606)

runningcode:
Extracts tooling version information (sentry-cli, Fastlane plugin, Gradle plugin) from uploaded artifact zip files and sends them to Sentry for database storage.

- Adds `_parse_metadata_content` function to parse the metadata file format:
  - `sentry-cli-version: X.Y.Z`
  - `sentry-fastlane-plugin: X.Y.Z`
  - `sentry-gradle-plugin: X.Y.Z`

Refs: EME-606

**Related PRs:**
getsentry/sentry#103062
https://github.com/getsentry/sentry/pull/104846
https://github.com/getsentry/sentry-cli/pull/2994

### DIFF
--- a/src/launchpad/api/update_api_models.py
+++ b/src/launchpad/api/update_api_models.py
@@ -37,7 +37,11 @@ PutSize = Annotated[
 ]
 
 
-class AppleAppInfo(BaseModel):
+class BaseAppInfo(BaseModel):
+    cli_version: Optional[str] = Field(None, description="sentry-cli version used for uploading")
+
+
+class AppleAppInfo(BaseAppInfo):
     is_simulator: bool
     codesigning_type: Optional[str] = None
     profile_name: Optional[str] = None
@@ -48,10 +52,12 @@ class AppleAppInfo(BaseModel):
     certificate_expiration_date: Optional[str] = None
     missing_dsym_binaries: Optional[List[str]] = None
     build_date: Optional[str] = None
+    fastlane_plugin_version: Optional[str] = Field(None, description="Fastlane plugin version used for uploading")
 
 
-class AndroidAppInfo(BaseModel):
+class AndroidAppInfo(BaseAppInfo):
     has_proguard_mapping: bool
+    gradle_plugin_version: Optional[str] = Field(None, description="Gradle plugin version used for uploading")
 
 
 class UpdateData(BaseModel):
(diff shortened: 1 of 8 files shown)


src/launchpad/utils/metadata_extractor.py
```diff
@@ -0,0 +1,90 @@
...
+
+def _parse_metadata_content(content: str) -> ToolingMetadata:
+    """Parse the content of .sentry-cli-metadata.txt file.
+
+    Expected format:
+        sentry-cli-version: 2.58.2
+        sentry-fastlane-plugin: 1.2.3
+        sentry-gradle-plugin: 4.12.0
+
+    Args:
```
rbro112: IMO these arg/return comments aren't useful, the format is useful IMO so you can leave that in

src/launchpad/artifact_processor.py
```diff
@@ -459,6 +459,9 @@ def _get_artifact_type(artifact: Artifact) -> ArtifactType:
             apple_app_info=apple_app_info,
             android_app_info=android_app_info,
             dequeued_at=dequeued_at,
+            cli_version=app_info.cli_version,
+            fastlane_plugin_version=app_info.fastlane_plugin_version,
```
rbro112: Not sure I see a point of setting these here if you're already setting them in the `apple_app_info`/`android_app_info` - I'd recommend only one spot or the other

rbro112: `apple_app_info` and `android_app_info` are already top-level fields in the `UpdateData` model. You also have reference to those in the backend. In this case, you're still duplicating the fields, `cli_version`, `fastlane_plugin_version` and `gradle_plugin_version`. Those can and should be pulled from only one spot or another (top-level in UpdateData or in apple/android_app_info), not duplicated.

IMO there should be a PR here to update the backend handling to get those values from the android/apple_app_info fields and we should remove the top-level fields from UpdateData. Simpler and leverages the app_info models that are already being passed, and we don't duplicate data.

src/launchpad/api/update_api_models.py
```diff
@@ -37,7 +37,11 @@ class PutSizePending(BaseModel):
 ]
 
 
-class AppleAppInfo(BaseModel):
+class BaseAppInfo(BaseModel):
```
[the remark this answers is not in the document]
rbro112: Nope, perfect use case of this base pattern 👍

src/launchpad/artifact_processor.py
```diff
@@ -442,12 +442,16 @@ def _get_artifact_type(artifact: Artifact) -> ArtifactType:
                 certificate_expiration_date=app_info.certificate_expiration_date,
                 missing_dsym_binaries=app_info.missing_dsym_binaries,
                 build_date=app_info.build_date,
+                cli_version=app_info.cli_version,
```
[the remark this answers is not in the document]
rbro112: I'd say keep here so that the apple_app_info/android_app_info keep parity with our backend models

PR #469, opened 2025-11-11
Add join_timeout value for kafka consumer to better handle rebalances

NicoHinderling:
A kafka rebalance was triggered (t=0) which caused JoinGroupRequest to be sent. This request has a 246-second timeout.

**t=0s - t=57s**: Consumer stuck in rebalance operation
Main thread blocked coordinating the rebalance
Heartbeat background thread couldn't send heartbeats (or they were queued/blocked)
Session timeout: 45 seconds
**t=45s**: Kafka broker says "No heartbeat for 45 seconds → consumer is dead"
Marks consumer as failed
But requests are still in-flight...
**t=57s**: Connection forcibly closed

Supposedly instead of bumping the `session.timeout.ms` value we should instead provide a `join_timeout` value so that our consumer will drop the work

### DIFF
--- a/src/launchpad/kafka.py
+++ b/src/launchpad/kafka.py
@@ -138,6 +138,7 @@ def create_kafka_consumer() -> LaunchpadKafkaConsumer:
         consumer=arroyo_consumer,
         topic=topic,
         processor_factory=strategy_factory,
+        join_timeout=config.join_timeout_seconds,  # Drop in-flight work during rebalance before Kafka times out
     )
     return LaunchpadKafkaConsumer(processor, healthcheck_path, strategy_factory)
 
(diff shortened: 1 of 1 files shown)


src/launchpad/kafka.py
```diff
@@ -111,6 +111,7 @@ def create_kafka_consumer() -> LaunchpadKafkaConsumer:
         "enable.auto.commit": False,
         "enable.auto.offset.store": False,
         "security.protocol": config.security_protocol,
+        "session.timeout.ms": 300000,  # 5 minutes - how long before Kafka considers consumer dead. Add enough time for worst case rebalance
```
trevor-e: Should this be routed through our `KafkaConfig`?

NicoHinderling: I was on the fence since this value was based on the rebalancing timeout, but I'll revisit and potentially move it after confirming with streaming infra whether this is a good solution

trevor-e (approved):

PR #473, opened 2025-11-14
Fix various iOS treemap inaccuracies 

trevor-e:
There were a couple bugs that I was finally able to track down. 

1. Not handling zerofill sections properly. These sections contain uninitialized symbols and don't take up disk space. The section has a size but that's only used when loading into virtual memory. If you summed up all the section sizes in lief they were greater than the segment size due to this bug.
2. Handling Swift symbols we can't attribute. Our custom demangling code seems to have a bug and can't demangle about ~1% of symbols, and we were silently dropping these during our calculation.
3. Moves unmapped nodes to be per segment.

There are some slight counting discrepancies but very small now, only `7368` over-counting bytes for the main Hackernews binary:
```
           WARNING  macho.treemap.size_mismatch | binary_name=HackerNews file_size=3153920 treemap_size=3161288 difference=7368 difference_type=over_counted  
                    percentage=0.23                                                                                                                           
           DEBUG    Building treemap for Common                                                                                                               
           WARNING  macho.treemap.size_mismatch | binary_name=Common file_size=192512 treemap_size=194536 difference=2024 difference_type=over_counted        
                    percentage=1.05                                                                                                                           
           DEBUG    Building treemap for Reaper                                                                                                               
           WARNING  macho.treemap.size_mismatch | binary_name=Reaper file_size=53248 treemap_size=35792 difference=17456 difference_type=missing              
                    percentage=32.78                                                                                                                          
           DEBUG    Building treemap for Sentry                                                                                                               
           WARNING  macho.treemap.size_mismatch | binary_name=Sentry file_size=53248 treemap_size=35792 difference=17456 difference_type=missing              
                    percentage=32.78                                                                                                                          
           DEBUG    Building treemap for HackerNewsHomeWidgetExtension                                                                                        
           WARNING  macho.treemap.size_mismatch | binary_name=HackerNewsHomeWidgetExtension file_size=155648 treemap_size=157672 difference=2024              
                    difference_type=over_counted percentage=1.3 
```

### DIFF
--- a/tests/unit/test_swift_symbol_type_aggregator.py
+++ b/tests/unit/test_swift_symbol_type_aggregator.py
@@ -30,12 +30,13 @@ class TestSwiftSymbolTypeAggregator:
         assert SwiftSymbolTypeAggregator.is_swift_symbol("_TtC6Sentry14OnDemandReplay")
         assert SwiftSymbolTypeAggregator.is_swift_symbol("_Tt")
 
-    def test_is_swift_symbol_objc_metadata(self):
-        """Test Swift symbol detection for ObjC runtime metadata."""
-        # Swift classes exposed to ObjC with metadata
-        assert SwiftSymbolTypeAggregator.is_swift_symbol("__IVARS__TtC6Sentry14OnDemandReplay")
-        assert SwiftSymbolTypeAggregator.is_swift_symbol("__DATA__TtC6Sentry14OnDemandReplay")
-        assert SwiftSymbolTypeAggregator.is_swift_symbol("_OBJC_CLASS_$__TtC6Sentry14OnDemandReplay")
+    # TODO(EME-645): bring back later
+    # def test_is_swift_symbol_objc_metadata(self):
+    #     """Test Swift symbol detection for ObjC runtime metadata."""
+    #     # Swift classes exposed to ObjC with metadata
+    #     assert SwiftSymbolTypeAggregator.is_swift_symbol("__IVARS__TtC6Sentry14OnDemandReplay")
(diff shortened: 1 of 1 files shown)


src/launchpad/size/analyzers/apple.py
```diff
@@ -564,7 +564,8 @@ def _extract_segments_info(self, binary: lief.MachO.Binary) -> List[SegmentInfo]
                     section_infos: List[SectionInfo] = []
                     for section in command.sections:
                         section_name = self._parse_lief_name(section.name)
-                        section_infos.append(SectionInfo(name=section_name, size=section.size))
+                        is_zerofill = section.type == lief.MachO.Section.TYPE.ZEROFILL
```
noahsmartin: A segment has a `vmsize` and a `filesize`. Some segments have vmsize > 0 but filesize == 0. I've never seen a case where the segment has a file size but some sections do not have a file size, but maybe it's possible. Just to be extra sure can we add a guard above the section loop to skip any segment with `filesize == 0` ?

src/launchpad/size/analyzers/apple.py
```diff
@@ -564,7 +564,8 @@ def _extract_segments_info(self, binary: lief.MachO.Binary) -> List[SegmentInfo]
...
+                        section_infos.append(SectionInfo(name=section_name, size=section.size, is_zerofill=is_zerofill))
```
noahsmartin: Can we just not append to section_infos at all if it's zero fill? That seems easier?

trevor-e: spoke offline, basically want to keep this data unfiltered and have downstream consumers handle filtering stuff out

noahsmartin (approved):

PR #476, opened 2025-11-14
Fix plist date parsing

trevor-e:


src/launchpad/size/analyzers/apple.py
```diff
@@ -259,6 +259,23 @@ def analyze(self, artifact: AppleArtifact) -> AppleAnalysisResults:
...
+            try:
+                dt = datetime.fromisoformat(date_value)
+                return dt.isoformat()
+            except (ValueError, AttributeError):
+                # If parsing fails, return the original string
+                # This is defensive - we don't want to break on unexpected formats
+                logger.debug(f"Could not parse date string: {date_value}")
+                return date_value
+
+        # isinstance(date_value, datetime)
```
rbro112: Accidental comment?

rbro112 (approved):

src/launchpad/size/analyzers/apple.py
trevor-e: no intentional, its part of the type flow

PR #479, opened 2025-11-18
Convert arroyo consumer back to RunTaskInThreads + manually managed subprocess

NicoHinderling:
Allow us have timeouts for processes

### DIFF
--- a/src/launchpad/kafka.py
+++ b/src/launchpad/kafka.py
@@ -59,7 +59,7 @@ def process_kafka_message_with_service(
     log_queue: multiprocessing.Queue[Any],
 ) -> Any:
     """Process a Kafka message by spawning a fresh subprocess with timeout protection."""
-    timeout = int(os.getenv("KAFKA_TASK_TIMEOUT_SECONDS", "720"))  # 12 minutes default
+    timeout = int(os.getenv("KAFKA_TASK_TIMEOUT_SECONDS", "3600"))  # 1 hour default
 
     try:
         decoded = PREPROD_ARTIFACT_SCHEMA.decode(msg.payload.value)


src/launchpad/kafka.py
```diff
@@ -41,51 +36,60 @@
...
-            signal.SIGCHLD,
-            lambda signum, frame: logger.debug(f"Worker process exited normally (SIGCHLD {signum})"),
-        )
 
+def process_kafka_message_with_service(
+    msg: Message[KafkaPayload],
+    log_queue: multiprocessing.Queue[Any],
+) -> Any:
+    """Process a Kafka message by spawning a fresh subprocess with timeout protection."""
+    timeout = int(os.getenv("KAFKA_TASK_TIMEOUT_SECONDS", "720"))  # 12 minutes default
```
trevor-e: Maybe we should default to `None` for now to help us roll this out? That way the timeout is decoupled from the RunTaskInThreads change.

src/launchpad/kafka.py
```diff
@@ -41,51 +36,60 @@
...
 
+    root_logger.addHandler(queue_handler)
+    root_logger.setLevel(logging.DEBUG)
 
-class LaunchpadRunTaskWithMultiprocessing(RunTaskWithMultiprocessing[TStrategyPayload, Any]):
-    """Tolerates child process exits from maxtasksperchild=1 by ignoring SIGCHLD."""
+    try:
+        ArtifactProcessor.process_message(decoded_message)
+    except Exception as e:
+        logger.exception(f"Error processing message in subprocess: {e}")
```
trevor-e: nit: you don't need to pass `e` in here

src/launchpad/kafka.py
```diff
@@ -41,51 +36,60 @@
...
+        logger.error(f"Task exceeded timeout of {timeout}s, killing process | artifact_id={artifact_id}")
+        process.terminate()
+        process.join(timeout=5)  # Give it 5s to terminate gracefully
+        if process.is_alive():
+            process.kill()
+            process.join()
+        return None  # type: ignore[return-value]
+
+    if process.exitcode != 0:
+        logger.error(f"Process exited with code {process.exitcode} | artifact_id={artifact_id}")
```
trevor-e: nit: same

trevor-e (commented):
What's the plan for deploying this? Will we let it sit in S4S for a bit first to see if all the telemetry still works correctly?

PR #482, opened 2025-11-19
[iOS] Refactor get_all_binary_paths + add lief cache

NicoHinderling:
Added a cache to avoid a second parse. It only caches a binary if it contains a dSYM, since the second parse only concerns itself with binaries containing dSYMs.

In the process, I also cleaned up the structure of get_all_binary_paths a bit

### DIFF
--- a/src/launchpad/artifacts/apple/zipped_xcarchive.py
+++ b/src/launchpad/artifacts/apple/zipped_xcarchive.py
@@ -361,6 +361,10 @@ class ZippedXCArchive(AppleArtifact):
 
         return binaries
 
+    def get_lief_cache(self) -> dict[Path, lief.MachO.FatBinary]:
+        """Get the LIEF cache of pre-parsed binaries"""
+        return self._lief_cache
+
     @sentry_sdk.trace
     def get_asset_catalog_details(self, relative_path: Path) -> List[AssetCatalogElement]:
         """Get the details of an asset catalog file (Assets.car) by returning the
--- a/src/launchpad/size/analyzers/apple.py
+++ b/src/launchpad/size/analyzers/apple.py
@@ -151,7 +151,7 @@ class AppleAppAnalyzer:
             binaries = artifact.get_all_binary_paths()
             logger.debug(f"Found {len(binaries)} binaries to analyze")
 
-            lief_cache = getattr(artifact, "_lief_cache", {})
+            lief_cache = artifact.get_lief_cache()
             for binary_info in binaries:
                 logger.info(
                     "size.apple.binary_analysis_started",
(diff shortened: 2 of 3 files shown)


src/launchpad/size/analyzers/apple.py
```diff
@@ -151,6 +151,7 @@ def analyze(self, artifact: AppleArtifact) -> AppleAnalysisResults:
             binaries = artifact.get_all_binary_paths()
             logger.debug(f"Found {len(binaries)} binaries to analyze")
 
+            lief_cache = getattr(artifact, "_lief_cache", {})
```
trevor-e: We should very rarely have to use `getattr()` for typed code that we own.

src/launchpad/artifacts/apple/zipped_xcarchive.py
```diff
@@ -494,27 +495,75 @@ def _parse_asset_element(self, item: dict[str, Any], parent_path: Path) -> Asset
...
                     if command.command == lief.MachO.LoadCommand.TYPE.UUID:
                         if isinstance(command, lief.MachO.UUIDCommand):
                             uuid_bytes = bytes(command.uuid)
                             uuid_obj = uuid.UUID(bytes=uuid_bytes)
-                            return str(uuid_obj).upper()
+                            extracted_uuid = str(uuid_obj).upper()
+                            break
+
+                if extracted_uuid is None:
+                    logger.debug(f"No UUID command found in binary: {binary_path}")
```
trevor-e: the old `def _extract_binary_uuid(self, binary_path: Path) -> str | None:` returns a nullable, we should still cache the parsed binary if no uuid is found

src/launchpad/artifacts/apple/zipped_xcarchive.py
```diff
@@ -494,27 +495,75 @@ def _parse_asset_element(self, item: dict[str, Any], parent_path: Path) -> Asset
...
+                    continue
+
+                self._binary_uuid_cache[binary_path] = extracted_uuid
+                if extracted_uuid in self._dsym_info:
```
trevor-e: Same with this check, not entirely sure what it's doing. We should cache the lief.parse() call in every case unless I'm misunderstanding something.

src/launchpad/size/analyzers/apple.py
```diff
@@ -448,9 +450,11 @@ def _analyze_binary(
 
         logger.debug(f"Analyzing binary: {binary_path}")
 
-        with open(binary_path, "rb") as f:
-            fat_binary = lief.MachO.parse(f)  # type: ignore
-
+        fat_binary = lief_cache.pop(binary_path, None) if lief_cache else None
```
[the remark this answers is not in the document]
NicoHinderling: I'll add an extra comment to make sure this is more obvious:

`# Cache is only used here currently. "get & remove" from cache to free up memory ASAP to minimize peak memory usage.`

PR #498, opened 2025-12-02
[S4S Testing] Get rid of healthcheck validation

NicoHinderling:
Validate whether pods being shut down during rebalance have to do with healthcheck indeed

### DIFF
--- a/src/launchpad/kafka.py
+++ b/src/launchpad/kafka.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import os
-import time
 
 from dataclasses import dataclass
 from typing import Any, Mapping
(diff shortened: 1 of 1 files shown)


src/launchpad/service.py
```diff
@@ -93,9 +93,7 @@ def signal_handler(signum: int, frame) -> None:
 
     def is_healthy(self) -> bool:
         """Get overall service health status."""
-        is_server_healthy = self.server.is_healthy()
-        is_kafka_healthy = self.kafka.is_healthy()
-        return is_server_healthy and is_kafka_healthy
```
trevor-e: Could you instead just change `kafka`'s `is_healthy` to always return true like `server.py` already does? That way we only have to touch one file and it's clear we are short-circuiting Kafka's health check.

NicoHinderling: sure

trevor-e (approved):

PR #512, opened 2025-12-05
Add is_shutting_down flag for arroyo rebalance case

NicoHinderling:
After merging https://github.com/getsentry/launchpad/pull/504 we noticed that there was still cases of the poll interval timeout happening. Upon further investigation, we realized this is happening:

```
Normal Processing:
  Message 1 → Worker spawns subprocess, waits
  Message 2 → Queued in ThreadPoolExecutor (concurrency=1, worker busy)

Rebalance:
  1. Kill Message 1's subprocess (~6s)
  2. Worker becomes free
  3. Executor auto-assigns Message 2 to freed worker
  4. NEW subprocess starts (11+ minutes)
  5. executor.shutdown() blocks waiting for Message 2
  6. Rebalance takes 11+ min → max.poll.interval.ms exceeded
  ```
  
  To validate this, I set `KAFKA_MAX_PENDING_FUTURES` to 0 and that killed the topic lag quickly on US region yesterday. This works because it means that arroyo's `run_task_in_threads` wont have any future tasks pre-fetched. Instead of doing that though, we can have a flag set when shut down is happening and then just skip any future tasks that attempt to get processed during the shutdown rebalance.
  
  This way we can keep `KAFKA_MAX_PENDING_FUTURES` non zero and still have the benefit of prefetching tasks (and potentially still our consumer with a concurrency > 1 if we needed to in the future

### DIFF
--- a/src/launchpad/kafka.py
+++ b/src/launchpad/kafka.py
@@ -56,14 +56,14 @@ def _process_in_subprocess(decoded_message: Any, log_queue: multiprocessing.Queu
 def _kill_process(process: multiprocessing.Process, artifact_id: str) -> None:
     """Gracefully terminate, then force kill a subprocess."""
     process.terminate()
-    process.join(timeout=5)
+    process.join(timeout=1)
     if process.is_alive():
         logger.warning(
             "Process did not terminate gracefully, force killing",
             extra={"artifact_id": artifact_id, "pid": process.pid},
         )
         process.kill()
-        process.join(timeout=1)  # Brief timeout to reap zombie, avoid infinite block
+        process.join(timeout=0.5)
         if process.is_alive():
             logger.error(
                 "Process could not be killed, may become zombie",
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/launchpad/kafka.py [hunk 1]
```diff
@@ -79,6 +79,9 @@ def process_kafka_message_with_service(
     factory: LaunchpadStrategyFactory,
 ) -> Any:
     """Process a Kafka message by spawning a fresh subprocess with timeout protection."""
+    if factory._is_shutting_down:
+        raise TimeoutError("Skipping message processing - shutdown in progress")
```
trevor-e: `TimeoutError` seems like a misleading error to raise in these cases, the processing itself isn't timing out at all.

src/launchpad/kafka.py
```diff
@@ -101,21 +104,12 @@ def process_kafka_message_with_service(
...
-                    factory._killed_during_rebalance.discard(pid)
-
-            if was_killed_by_rebalance:
-                # Wait for kill to complete, then don't commit offset
-                process.join(timeout=10)  # Give kill_active_processes time to finish
-                logger.warning(
-                    "Process killed during rebalance, message will be reprocessed",
-                    extra={"artifact_id": artifact_id},
-                )
-                raise TimeoutError("Subprocess killed during rebalance")
```
trevor-e: Although I realize we were already using `TimeoutError` here.

NicoHinderling:
Going with https://github.com/getsentry/launchpad/pull/513 instead for now, but keeping this PR in our back pocket for the reasons cited in 513's summary

PR #527, opened 2025-12-15
Add ability to unzip artifacts zipped with zstd

NicoHinderling:
Partner PR to https://github.com/getsentry/sentry-cli/pull/3038

Noticed uploading a certain customer's build took 5+ minutes and upon further inspection the zipping was the lions share of that time. If we take away the actual network transport of the upload:

With level 6 compression: 5 minutes -> ultimately 566MB
with level 1 compression: 2.5 min -> ultimately 658MB
with zstandard: 36s -> ultimately 533MB

This change will allow us to support both types of compression on the Sentry backend

### DIFF
--- a/src/launchpad/__init__.py
+++ b/src/launchpad/__init__.py
@@ -1,4 +1,4 @@
 # Monkey patches - import these first to register handlers globally
-import zipfile_zstd  # noqa: F401 - Registers zstd compression support with zipfile module
+import zipfile_zstd  # noqa: F401 - Registers zstd compression support with zipfile module. Should not be required after upgrading to python 3.14
 
-__version__ = "0fg.0.1"
+__version__ = "0.0.1"
--- a/tests/unit/artifacts/providers/test_zip_provider.py
+++ b/tests/unit/artifacts/providers/test_zip_provider.py
@@ -128,7 +128,7 @@ class TestCheckReasonableZip:
 
     def test_extract_zstd_zip(self) -> None:
         """Test that zstd-compressed zips can be extracted."""
-        with tempfile.NamedTemporaryFile(suffix=".zip", delete=False) as temp_file:
+        with tempfile.NamedTemporaryFile(suffix=".zip") as temp_file:
             temp_path = Path(temp_file.name)
 
             # Create a zstd-compressed zip (compression method 93)


tests/unit/artifacts/providers/test_zip_provider.py
```diff
@@ -125,3 +125,22 @@ def test_max_file_size(self, hackernews_xcarchive: Path) -> None:
...
                 check_reasonable_zip(zf, max_uncompressed_size=10 * 1024 * 1024)
+
+    def test_extract_zstd_zip(self) -> None:
+        """Test that zstd-compressed zips can be extracted."""
+        with tempfile.NamedTemporaryFile(suffix=".zip", delete=False) as temp_file:
+            temp_path = Path(temp_file.name)
+
+            # Create a zstd-compressed zip (compression method 93)
+            with zipfile.ZipFile(temp_path, "w") as zf:
+                zf.writestr("test.txt", "content", compress_type=93)
```
trevor-e: nit: can this use `zipfile.ZIP_ZSTANDARD` instead of magic number 93?

src/launchpad/__init__.py
```diff
@@ -1 +1,4 @@
-__version__ = "0.0.1"
+# Monkey patches - import these first to register handlers globally
+import zipfile_zstd  # noqa: F401 - Registers zstd compression support with zipfile module
+
+__version__ = "0fg.0.1"
```
trevor-e: Why the version bump?

requirements.txt
```diff
@@ -20,3 +20,4 @@ sentry-kafka-schemas==2.1.2
 sentry-sdk>=2.36.0
 sortedcontainers>=2.4.0
 typing-extensions>=4.15.0
+zipfile-zstd==0.0.4
```
trevor-e: It looks like we get zstd for free in Python 3.14, not sure if you looked into that option: https://docs.python.org/3/library/compression.zstd.html#

NicoHinderling: i tried upgrading us to 3.14, but that requires upgrading `confluent-kafka` to 2.12 which then conflicts with `sentry-arroyo` which has that package pinned to <2.10.0... so not possible right now

hen we're off of arroyo though.. lol

PR #543, opened 2026-01-21
Update DuplicateFiles Insight to better support directory duplicates

NicoHinderling:
Resolves EME-740

Duplicate directory detection was reporting ~4KB savings instead of full content size (e.g., 60MB bundles showed 4KB savings). This is because `FileInfo.size` for directories only stores the directory entry size, not the total content size. The full size was computed during directory hashing but never stored.

So I added `size_including_children` to the FileInfo model which we now use in `duplicate_files.py`

Also I've made sure that we won't report files as duplicates in directories which we've reported as duplicates as well (aka no double counting)

### DIFF
--- a/src/launchpad/size/insights/common/duplicate_files.py
+++ b/src/launchpad/size/insights/common/duplicate_files.py
@@ -42,20 +42,16 @@ class DuplicateFilesInsight(Insight[DuplicateFilesInsightResult]):
                 continue
 
             # Savings if we keep the largest one and dedupe the rest
-            # size_including_children is always set for directories
-            dirs.sort(key=lambda d: (-d.size_including_children, d.path))  # type: ignore[arg-type]
+            dirs.sort(key=lambda d: (-self._dir_size(d), d.path))
             if len(dirs) < 2:
                 continue
 
-            group_size = sum(d.size_including_children for d in dirs)  # type: ignore[misc]
-            savings = group_size - dirs[0].size_including_children  # type: ignore[operator]
+            group_size = sum(self._dir_size(d) for d in dirs)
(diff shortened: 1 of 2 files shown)


tests/unit/size/test_duplicate_files_insight.py
```diff
@@ -134,6 +134,7 @@ def test_generate_with_duplicate_directories_prevents_double_counting(self):
                 treemap_type=TreemapType.OTHER,
                 hash="bundle_hash",  # Same hash for identical content
                 is_dir=True,
+                size_including_children=12000,
```
trevor-e: Can we add some extra asserts for this `test_generate_with_duplicate_directories_prevents_double_counting` test to confirm the duplicate file insight size? It seems like this should have caught the issue.

src/launchpad/size/insights/common/duplicate_files.py
```diff
@@ -42,16 +42,20 @@ def generate(self, input: InsightsInput) -> DuplicateFilesInsightResult | None:
...
             # Savings if we keep the largest one and dedupe the rest
-            dirs.sort(key=lambda d: (-d.size, d.path))
+            # size_including_children is always set for directories
+            dirs.sort(key=lambda d: (-d.size_including_children, d.path))  # type: ignore[arg-type]
             if len(dirs) < 2:
                 continue
 
-            group_size = sum(d.size for d in dirs)
-            savings = group_size - dirs[0].size
+            group_size = sum(d.size_including_children for d in dirs)  # type: ignore[misc]
```
trevor-e: nit: are all the `# type: ignore` needed? the type you added for `size_including_children` looks pretty normal to me

PR #553, opened 2026-01-29
Add `.devenv/bin` to PATH

george-sentry:
The `.devenv/all/bin` directory is not created on my machine, so `devenv sync` fails. For some reason, I get a `.devenv/bin` directory instead, so I added it to PATH in case someone else has a similar setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ devservices up
 launchpad serve
 ```
 
-And finally use the `sentry-cli` to upload to your local machine:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -101,7 +101,7 @@ devservices up
 launchpad serve
 ```
 
-And finally use the `sentry-cli` to upload to your local machine:
+And finally use the `sentry-cli` (version 3.10 or greater) to upload to your local machine:
```
NicoHinderling: should this be 3.1.0?

NicoHinderling (approved):

PR #558, opened 2026-01-30
fix: check segment name when matching symbol neighbors

trevor-e:
We had a bug where in some rare cases we would compare a symbol to a neighbor in a different segment, which in this case produced a negative symbol size and caused a pydantic validation error. Now we ensure only symbols within the same segment are compared since some sections can have colliding names.

Tested the affected customer build and our hackernews fixture and seems correct.

https://sentry.sentry.io/issues/7215094878/events/c311e7b4e31a4ff98783579d17c581b9/#logs

### DIFF
--- a/src/launchpad/size/symbols/macho_symbol_sizes.py
+++ b/src/launchpad/size/symbols/macho_symbol_sizes.py
@@ -3,10 +3,16 @@ from dataclasses import dataclass
 from typing import NamedTuple
 
 import lief
+import sentry_sdk
 
 from launchpad.utils.logging import get_logger
 
 logger = get_logger(__name__)
+
+
+def _decode_name(raw_name: str | bytes) -> str:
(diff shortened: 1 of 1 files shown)


src/launchpad/size/symbols/macho_symbol_sizes.py
```diff
@@ -34,19 +40,16 @@ def __init__(self, binary: lief.MachO.Binary) -> None:
...
-        for data in symbol_data:
-            symbol_sizes.append(
-                SymbolSize(
-                    mangled_name=data.name,
-                    section_name=data.section_name,
-                    segment_name=data.segment_name,
-                    address=data.address,
-                    size=data.size,
-                )
+        symbol_sizes = [
```
trevor-e: There's some light cleanup mixed in here as well, left a comment on the bug fix.

PR #567, opened 2026-02-23
fix(artifacts): Report distribution errors for unsupported and skipped builds (EME-422)

runningcode:
## Summary

When Sentry requests `BUILD_DISTRIBUTION`, Launchpad now reports back via
the distribution endpoint in all cases where it can't fulfill the request:

- **Unknown artifact types**: Reports `PROCESSING_ERROR` via `update_distribution_error`
- **Invalid code signature (iOS)**: Reports `SKIPPED` with reason `invalid_signature`
- **Simulator builds (iOS)**: Reports `SKIPPED` with reason `simulator`

Previously, these cases either silently did nothing or only logged an error,
leaving the artifact stuck in a "processing" state in Sentry with no timeout
to recover.

Fixes EME-422

### DIFF
--- a/src/launchpad/artifact_processor.py
+++ b/src/launchpad/artifact_processor.py
@@ -370,7 +370,7 @@ class ArtifactProcessor:
                     error_code=InstallableAppErrorCode.PROCESSING_ERROR.value,
                     error_message="unsupported artifact type",
                 )
-            except SentryClientError:
+            except Exception:
                 logger.exception(f"Failed to update distribution error for artifact {artifact_id}")
 
     def _do_size(
@@ -468,7 +468,7 @@ class ArtifactProcessor:
                 error_code=InstallableAppErrorCode.SKIPPED.value,
                 error_message=skip_reason,
             )
-        except SentryClientError:
+        except Exception:
             logger.exception(f"Failed to update distribution skip for artifact {artifact_id}")
 
     def _update_size_error_from_exception(
--- a/src/launchpad/artifact_processor.py
+++ b/src/launchpad/artifact_processor.py
@@ -330,13 +330,32 @@ class ArtifactProcessor:
         logger.info(f"BUILD_DISTRIBUTION for {artifact_id} (project: {project_id}, org: {organization_id})")
         if isinstance(artifact, ZippedXCArchive):
             apple_info = cast(AppleAppInfo, info)
-            if apple_info.is_code_signature_valid and not apple_info.is_simulator:
-                with tempfile.TemporaryDirectory() as temp_dir_str:
-                    temp_dir = Path(temp_dir_str)
-                    ipa_path = temp_dir / "App.ipa"
-                    artifact.generate_ipa(ipa_path)
-                    with open(ipa_path, "rb") as f:
-                        self._sentry_client.upload_installable_app(organization_id, project_id, artifact_id, f)
+            if not apple_info.is_code_signature_valid:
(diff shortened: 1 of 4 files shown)


src/launchpad/artifact_processor.py
```diff
@@ -438,6 +453,23 @@ def _update_artifact_error(
...
+    ) -> None:
+        """Report distribution skip via the dedicated distribution endpoint."""
+        try:
+            self._sentry_client.update_distribution_error(
+                org=organization_id,
+                artifact_id=artifact_id,
+                error_code=InstallableAppErrorCode.SKIPPED.value,
+                error_message=skip_reason,
+            )
+        except Exception:
```
chromy: Maybe this should more closely mirror `_update_artifact_error` above - that way it could cover both the skip and the error cases - and report the skipped / distro errors to `_statsd`

src/launchpad/sentry_client.py
```diff
@@ -252,6 +252,24 @@ def update_artifact(self, org: str, project: str, artifact_id: str, data: Dict[s
         endpoint = f"/api/0/internal/{org}/{project}/files/preprodartifacts/{artifact_id}/update/"
         return self._make_json_request("PUT", endpoint, UpdateResponse, data=data)
 
+    def update_distribution_error(self, org: str, artifact_id: str, error_code: int, error_message: str) -> None:
+        """Report distribution error via the dedicated distribution endpoint."""
+        endpoint = f"/api/0/organizations/{org}/preprodartifacts/{artifact_id}/distribution/"
+        url = self._build_url(endpoint)
+        body = json.dumps({"error_code": error_code, "error_message": error_message}).encode("utf-8")
+
+        logger.debug(f"PUT {url}")
```
chromy: nit: remove

src/launchpad/sentry_client.py
```diff
@@ -252,6 +252,24 @@ def update_artifact(self, org: str, project: str, artifact_id: str, data: Dict[s
...
+        response = self.session.request(
+            method="PUT",
+            url=url,
+            data=body,
+            auth=self.auth,
+            timeout=30,
+        )
+
+        if response.status_code != 200:
+            raise SentryClientError(response=response)
```
chromy: Can this not use `_make_json_request`?

src/launchpad/sentry_client.py
```diff
@@ -252,6 +252,24 @@ def update_artifact(self, org: str, project: str, artifact_id: str, data: Dict[s
         endpoint = f"/api/0/internal/{org}/{project}/files/preprodartifacts/{artifact_id}/update/"
         return self._make_json_request("PUT", endpoint, UpdateResponse, data=data)
 
+    def update_distribution_error(self, org: str, artifact_id: str, error_code: int, error_message: str) -> None:
```
chromy: I would be tempted to make this like `update_artifact` above (both naming and implementation wise)

src/launchpad/artifact_processor.py
runningcode: good point, added statsd and made the method look more similar to the above

src/launchpad/sentry_client.py
runningcode: Good idea, updated. I had to change the `_update_resonse_model` method so that we don't return a ResponseModel since this is fire and forget.

chromy (commented):
Lgtm

chromy (approved):
lgtm

PR #595, opened 2026-04-03
feat(worker): Make max_child_task_count configurable via environment variable

NicoHinderling:
Make `max_child_task_count` configurable via the `LAUNCHPAD_WORKER_MAX_CHILD_TASK_COUNT`
environment variable instead of being hardcoded to 1000.

The previous hardcoded value of 1000 was too high for our current workload and caused
resource issues. The new default is 1, which can be overridden per-environment as needed.

Changes:
- Add `max_child_task_count` field to `WorkerConfig` dataclass
- Read from `LAUNCHPAD_WORKER_MAX_CHILD_TASK_COUNT` env var with default of 1
- Pass the configured value to `TaskWorker` instead of the hardcoded 1000


src/launchpad/worker/config.py
```diff
@@ -55,7 +65,7 @@ def run_worker(processing_pool_name: str = "launchpad") -> None:
     worker = TaskWorker(
         app_module="launchpad.worker.app:app",
         broker_hosts=config.rpc_hosts,
-        max_child_task_count=1000,
+        max_child_task_count=config.max_child_task_count,
         concurrency=config.concurrency,
         child_tasks_queue_maxsize=config.concurrency * 2,
```
trevor-e: Is this saying that workers can queue up tasks as well? Do we want that behavior?

trevor-e (approved):

src/launchpad/worker/config.py
NicoHinderling: good catch, disabling it

PR #598, opened 2026-04-10
feat: Log Objectstore config on Client creation

lcian:
In an effort to debug the auth failures we're still seeing from the taskworkers in prod, add a log to see how the Objectstore client is being created.


chromy (approved):
lgtm

PR #603, opened 2026-04-23
feat(artifacts): Emit granular installable app error codes (EME-883)

runningcode:
Mirror the new granular `InstallableAppErrorCode` values added in [getsentry/sentry#113440](https://github.com/getsentry/sentry/pull/113440), and emit them directly from `_do_distribution` instead of packing the reason into the free-form `error_message` string.

Before, when distribution was skipped for one of these reasons we sent `error_code=SKIPPED` (or `PROCESSING_ERROR`) plus a short-code string in `error_message` and the Sentry frontend had to pattern-match on those strings. After Sentry gained dedicated enum values for each case, launchpad emits the specific code paired with a human-readable sentence that the Sentry install page renders verbatim as the description:

| Case                        | Before                                        | After                                                                           |
| --------------------------- | --------------------------------------------- | ------------------------------------------------------------------------------- |
| iOS signature invalid       | `SKIPPED` + `"invalid_signature"`             | `INVALID_CODE_SIGNATURE` + `"The build's code signature could not be verified."` |
| iOS simulator build         | `SKIPPED` + `"simulator"`                     | `SIMULATOR_BUILD` + `"Simulator builds cannot be distributed."`                 |
| Unrecognised artifact type  | `PROCESSING_ERROR` + `"unsupported artifact type"` | `UNSUPPORTED_ARTIFACT_TYPE` + `"This artifact type is not supported for distribution."` |

## Cleanup

- Drop the `error_message` tag from the `distribution.processing.error` statsd metric. `error_code` is the pivot dimension; `error_message` is 1:1 with the code today and interpolating a variable into it later would silently explode metric cardinality. The log line and PUT body still carry the message.
- Correct the class-name references in the enum sync comments (`InstallableApp.ErrorCode` → `PreprodArtifact.InstallableAppErrorCode`) and spell out why the invariant matters — drift will 400 on Sentry's pydantic validation.

Refs EME-883

### DIFF
--- a/src/launchpad/artifact_processor.py
+++ b/src/launchpad/artifact_processor.py
@@ -334,13 +334,13 @@ class ArtifactProcessor:
             if not apple_info.is_code_signature_valid:
                 logger.warning(f"BUILD_DISTRIBUTION skipped for {artifact_id}: invalid code signature")
                 self._update_distribution_error(
-                    organization_id, artifact_id, InstallableAppErrorCode.SKIPPED, "invalid_signature"
+                    organization_id, artifact_id, InstallableAppErrorCode.INVALID_CODE_SIGNATURE, ""
                 )
                 return
             if apple_info.is_simulator:
                 logger.warning(f"BUILD_DISTRIBUTION skipped for {artifact_id}: simulator build")
                 self._update_distribution_error(
-                    organization_id, artifact_id, InstallableAppErrorCode.SKIPPED, "simulator"
+                    organization_id, artifact_id, InstallableAppErrorCode.SIMULATOR_BUILD, ""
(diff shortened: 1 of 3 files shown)


src/launchpad/artifact_processor.py
```diff
@@ -465,7 +474,6 @@ def _update_distribution_error(
             "distribution.processing.error",
             tags=[
                 f"error_code:{error_code.value}",
-                f"error_message:{error_message}",
```
runningcode, line 468: I removed this since the error codes now map 1:1 with error messages.

chromy (approved):
lgtm

PR #608, opened 2026-04-29
fix(workflows): Drop injectable args, let sentry-cli auto-detect VCS

runningcode:
## Summary

Alternate fix to #607 for the script injection vulnerability in the Sentry Dogfood workflow.

Rather than moving every `${{ github.* }}` interpolation into an `env:` block, this PR removes the arguments entirely. `sentry-cli`'s `collect_git_metadata` auto-detects all of them when running in CI.

This eliminates the injection vector by construction — there is no untrusted input in any `run:` block — and collapses the workflow from ~95 lines of conditional branching to ~14 lines of straight-line invocation.

## Related

- Original fix: #607
- Linear: [VULN-1590](https://linear.app/getsentry/issue/VULN-1590)

### DIFF
--- a/.github/workflows/sentry_dogfood.yml
+++ b/.github/workflows/sentry_dogfood.yml
@@ -21,10 +21,7 @@ jobs:
           fetch-depth: 0
 
       - name: Install Sentry CLI
-        run: |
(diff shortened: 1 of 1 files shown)


cameroncooke (approved):
Much better, thanks @runningcode

PR #614, opened 2026-05-08
feat(insights): Add timeout to image optimization analysis

NicoHinderling:
Add a 300-second deadline to `BaseImageOptimizationInsight.generate()` that
stops processing images when the timeout is reached. This prevents the image
optimization insight from blocking the entire analysis pipeline for several
minutes on large apps with thousands of images (e.g. 3000+ asset catalog entries).

When the deadline fires:
- Pending futures are cancelled so queued work doesn't start
- A `timed_out` flag is set on `ImageOptimizationInsightResult` so consumers
  can distinguish "no optimizable images" from "analysis was cut short"
- Partial results collected before the timeout are still returned
- An empty result with `timed_out=True` is returned if no images passed the
  savings threshold before the deadline (previously this silently returned `None`)

The `timed_out` field defaults to `False` on the model so existing serialized
data deserializes correctly.

### DIFF
--- a/src/launchpad/size/insights/apple/image_optimization.py
+++ b/src/launchpad/size/insights/apple/image_optimization.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import io
 import logging
+import time
 
 from abc import ABC, abstractmethod
 from concurrent.futures import ThreadPoolExecutor, as_completed
@@ -46,6 +47,7 @@ class BaseImageOptimizationInsight(Insight[ImageOptimizationInsightResult], ABC)
     TARGET_JPEG_QUALITY = 85
     TARGET_HEIC_QUALITY = 85
     _MAX_WORKERS = 4
+    _TIMEOUT_SECONDS = 300
 
     @abstractmethod
     def _find_images(self, input: InsightsInput) -> List[FileInfo]:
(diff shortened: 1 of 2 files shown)


src/launchpad/size/insights/apple/image_optimization.py
```diff
@@ -73,17 +75,32 @@ def generate(self, input: InsightsInput) -> ImageOptimizationInsightResult | Non
             return None
 
         results: List[OptimizableImageFile] = []
+        timed_out = False
+        completed = 0
```
trevor-e, line 79: nit: is `completed` helpful here vs `len(results)`?

trevor-e (approved):
+ cursors comment

src/launchpad/size/insights/apple/image_optimization.py
NicoHinderling: iirc The `completed` counter tracks all processed futures (including failures and below-threshold images), while `len(results)` only counts optimization candidates

src/launchpad/size/insights/apple/image_optimization.py
```diff
@@ -73,17 +75,32 @@ def generate(self, input: InsightsInput) -> ImageOptimizationInsightResult | Non
...
+                    timed_out = True
+                    logger.warning(
+                        "size.insight.image_optimization.timeout | completed=%d/%d timeout=%ds",
+                        completed,
+                        len(files),
+                        self._TIMEOUT_SECONDS,
+                    )
+                    for fut in future_to_file:
+                        fut.cancel()
+                    break
```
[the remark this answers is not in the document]
NicoHinderling: Fixed — added a `completed < len(files)` guard before the deadline check so the `timed_out` flag is only set when work was actually skipped.

- Claude Code

PR #622, opened 2026-05-15
fix(security): Add path traversal guards to icon file reading

runningcode:
## Summary
- Zip extraction already validates paths via `is_safe_path()`, but after extraction, icon reading code constructed file paths from attacker-controlled manifest/plist data without checking for directory traversal
- Adds `is_safe_path()` checks to APK, AAB, and iOS icon reading, plus the icon parser's `_find_file()` method

Fixes EME-1152


chromy (approved):
lgtm

PR #623, opened 2026-05-15
ref(security): Add SafeDirectory to enforce path traversal checks

runningcode:
## Summary

Introduces `SafeDirectory` — a `Path`-like wrapper that validates untrusted input stays within bounds — and removes `is_safe_path()` entirely.

- `ZipProvider.extract_to_temp_directory()` returns `SafeDirectory` instead of `Path`
- `SafeDirectory.resolve(untrusted)` replaces all manual `is_safe_path()` + `raise UnsafePathError` patterns — developers can't accidentally skip the check
- `SafeDirectory.child(untrusted)` creates a scoped subdirectory (e.g. AAB's `base/` dir)
- Trusted `Path` operations (`glob`, `rglob`, `/`, `exists`, `relative_to`, etc.) are delegated directly, so `SafeDirectory` is a drop-in wherever `Path` was used before
- Icon parsers (`IconParser`, `BinaryXmlDrawableParser`, `ProtoXmlDrawableParser`) now accept only `SafeDirectory`, enforcing the safety boundary at the type level
- `UnsafePathError` / `UnreasonableZipError` moved to `exceptions.py`; all imports updated to use the canonical location
- `is_safe_path()` removed — `SafeDirectory.resolve()` is the single path validation mechanism
- Added SafeDirectory guidance to CLAUDE.md for future development

Fixes EME-1158

### DIFF
--- a/src/launchpad/artifacts/apple/zipped_xcarchive.py
+++ b/src/launchpad/artifacts/apple/zipped_xcarchive.py
@@ -131,7 +131,10 @@ class ZippedXCArchive(AppleArtifact):
         app_bundle = self.get_app_bundle_path()
 
         for icon_name in icon_info.primary_icon_files:
-            # Validate the plist-supplied name before globbing with it
+            if any(c in icon_name for c in ("*", "?", "[", "]")):
+                logger.warning(f"Skipping icon name with glob metacharacters: {icon_name}")
+                continue
+
             app_bundle.resolve(icon_name)
 
             # iOS lists base names without extensions or resolution modifiers (@2x, @3x, ~ipad)
--- a/src/launchpad/artifacts/apple/zipped_xcarchive.py
+++ b/src/launchpad/artifacts/apple/zipped_xcarchive.py
@@ -67,7 +67,7 @@ class ZippedXCArchive(AppleArtifact):
         super().__init__(path)
         self._zip_provider = ZipProvider(path)
         self._extract_dir = self._zip_provider.extract_to_temp_directory()
-        self._app_bundle_path: Path | None = None
+        self._app_bundle_path: SafeDirectory | None = None
         self._plist: dict[str, Any] | None = None
         self._archive_plist: dict[str, Any] | None = None
         self._provisioning_profile: dict[str, Any] | None = None
@@ -128,7 +128,7 @@ class ZippedXCArchive(AppleArtifact):
             logger.warning("No icon files found in CFBundleIconFiles")
             return None
 
-        app_bundle = SafeDirectory(self.get_app_bundle_path())
+        app_bundle = self.get_app_bundle_path()
 
         for icon_name in icon_info.primary_icon_files:
             # Validate the plist-supplied name before globbing with it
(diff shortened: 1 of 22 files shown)


src/launchpad/artifacts/providers/exceptions.py
```diff
@@ -0,0 +1,10 @@
+class UnreasonableZipError(ValueError):
```
runningcode, line 1: moved from zipprovider to avoid circular imports.

src/launchpad/artifacts/apple/zipped_xcarchive.py
```diff
@@ -127,16 +128,15 @@ def get_app_icon(self) -> bytes | None:
             logger.warning("No icon files found in CFBundleIconFiles")
             return None
 
-        app_bundle_path = self.get_app_bundle_path()
+        app_bundle = SafeDirectory(self.get_app_bundle_path())
 
         for icon_name in icon_info.primary_icon_files:
-            if not is_safe_path(app_bundle_path, icon_name):
-                raise UnsafePathError(f"Unsafe icon name in plist: {icon_name}")
+            app_bundle.resolve(icon_name)
```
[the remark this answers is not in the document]
cameroncooke: Agree, though should the "Trusted Path operations" filter their results as well? glob follows symlinks, so a result can dereference outside the base.

src/launchpad/artifacts/apple/zipped_xcarchive.py
```diff
@@ -127,16 +128,15 @@ def get_app_icon(self) -> bytes | None:
...
 
             # iOS lists base names without extensions or resolution modifiers (@2x, @3x, ~ipad)
             # Search for files matching the base name with any suffix
             # e.g., "AppIcon60x60" matches "AppIcon60x60@2x.png" or "AppIcon60x60.png"
-            matching_files = list(app_bundle_path.glob(f"{icon_name}*.png"))
+            matching_files = list(app_bundle.glob(f"{icon_name}*.png"))
```
cameroncooke, line 139: A malicious plist can put glob metacharacters in CFBundleIconFiles. resolve(icon_name) doesn't catch them because *, ?, [, **/ are all legal filename characters and the resolved path stays inside the base.

 Examples:
 - icon_name = "*" → glob("**.png") matches every PNG in the bundle root.
 - icon_name = "**/" → glob("**/*.png") matches every PNG anywhere in the bundle (** activates as recursive when it's a path segment).

While not exploitable it's probably something we want to tighten up on.

runningcode: Is that something introduced/changed in this PR?

src/launchpad/artifacts/apple/zipped_xcarchive.py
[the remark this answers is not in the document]
runningcode: Yeah it is needed. I think I'll just add a comment here to address the concerns.

src/launchpad/artifacts/apple/zipped_xcarchive.py
cameroncooke: No it's just something Claude pointed out as a vulnerability that we didn't close, though worse case is it just misdirects to another .png in same zip so not high risk, it can't leak out of the zip folder.

cameroncooke (approved):

PR #631, opened 2026-06-04
ref: brew-managed uv, requirements -> uv.lock, pin 3.14.4-slim-bookworm

kenzoengineer:
like here, we shouldn't manage UV anymore with devenv: https://github.com/getsentry/sentry-options-automator/pull/7343


devenv/config.ini
joshuarli: this should be removed too

devenv/sync.py
```diff
@@ -1,22 +1,23 @@
...
+        os.remove(f"{reporoot}/.devenv/bin/uv")
+
+    if os.path.exists(f"{reporoot}/.devenv/bin/uvx"):
+        os.remove(f"{reporoot}/.devenv/bin/uvx")
+
+    if not shutil.which("uv"):
+        print("\n\n\ndevenv is no longer managing uv; please run `brew install uv`.\n\n\n")
+        return 1
 
     venv_dir, python_version, requirements, editable_paths, bins = venv.get(reporoot, "launchpad")  # type: ignore
```
joshuarli: this can be replaced with

```
print("syncing .venv ...")
proc.run(
    ("uv", "sync", "--frozen", "--quiet", "--active"),
)
```

joshuarli (changes requested):
need to also add `brew 'uv'` to Brewfile and add

```
    if constants.DARWIN:
        brew.install()

        proc.run(
            (f"{constants.homebrew_bin}/brew", "bundle"),
            cwd=reporoot,
        )
```

joshuarli (approved):

PR #639, opened 2026-06-29
fix: Support push-based workers in launchpad

evanh:
The newer versions of the taskworker can be set to have tasks pushed to them instead of pulling activations from the broker. Update the CLI to allow those settings to be passed to the workers.


markstory (approved):
Looks good to me. The bot comments on config validation seem worth addressing for consistency.

PR #645, opened 2026-07-08
fix(preprod): Parse dotted Apple build numbers instead of dropping them

trevor-e:
## Summary
- `app_info.build` (CFBundleVersion) is only guaranteed to be a plain integer for Android's `versionCode`. Apple allows up to three dot-separated non-negative integers (e.g. `"1.2.3"`), which the previous `isdigit()` check silently turned into `None`.
- That broke Sentry's `build_number` tiebreaker used to sort/find the latest build within a version, and also made those artifacts non-installable on the Sentry side (worked around there separately).
- Pack multi-component builds into a single sortable int by zero-padding each component to a fixed width; the existing plain-integer path (Android, and simple Apple builds) is unchanged.
- Also send the raw, unparsed build string as `build_number_raw` so Sentry can store/display the exact value later without another launchpad deploy. Sentry currently ignores unrecognized fields, so this is safe to ship ahead of that follow-up work.
- Adds unit tests for `_parse_build_number` (plain ints unchanged, dotted values packed and correctly ordered, malformed input falls back to `None`) and for `_prepare_update_data`'s wiring of `build_number`/`build_number_raw`.

**Known limitation** (flagged in review): plain-integer build numbers stay small while packed dotted values are much larger, so if the same `app_id`/`build_version` ever has builds in both conventions, `build_number` alone doesn't reliably order them. This can't be resolved in launchpad — a single artifact update has no visibility into sibling artifacts' formats, and unifying the magnitude would break backward compatibility with every already-stored plain-integer `build_number`. The authoritative fix belongs in Sentry's tiebreak query once `build_number_raw` is available there (comparing the raw string directly for mixed-format cases), which is being scoped as a follow-up. Documented inline on `_parse_build_number`.

## Test plan
- [x] `make test-unit` (3 pre-existing, unrelated failures in Android APK signing/Swift symbol tests — confirmed present without this change)
- [ ] `make check`

### DIFF
--- a/src/launchpad/artifact_processor.py
+++ b/src/launchpad/artifact_processor.py
@@ -680,6 +680,16 @@ def _parse_build_number(build: str) -> int | None:
     zero-padding each component, which preserves correct ordering as long as no
     component reaches 10**_BUILD_NUMBER_COMPONENT_WIDTH. Anything else
     (non-numeric, malformed) returns None, same as today.
+
+    Known limitation: plain-integer values are left small while packed dotted
+    values are much larger, so if the same app_id/build_version ever has builds
+    in both conventions, this int alone is not a reliable ordering between them.
+    We can't fix that here — a single artifact update has no visibility into
+    sibling artifacts' formats, and unifying the magnitude would break backward
+    compatibility with every already-stored plain-integer build_number. Sentry's
+    tiebreak query should treat this as a fast, common-case sort key and fall
+    back to comparing the raw build string (build_number_raw) directly when it
+    needs to be authoritative across mixed formats.
     """
     if build.isdigit():
         return int(build)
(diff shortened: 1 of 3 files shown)


src/launchpad/artifact_processor.py
```diff
@@ -662,3 +663,24 @@ def _guess_message(code: ProcessingErrorCode, e: Exception) -> ProcessingErrorMe
...
+
+
+def _parse_build_number(build: str, component_width: int = 6) -> int | None:
+    """Parse a raw build identifier (e.g. CFBundleVersion) into a sortable int.
+
+    Plain integer builds (the common case, e.g. Android versionCode) pass through
+    unchanged. Apple's CFBundleVersion also allows up to three dot-separated
+    non-negative integers (e.g. "1.2.3"); those are packed into a single int by
+    zero-padding each component to `component_width` digits, which preserves
+    correct ordering as long as no component reaches 10**component_width.
```
trevor-e: This field is stored as a `BoundedBigInt` in Sentry so it should be safe to pad like this.

mtopo27 (approved):

PR #650, opened 2026-08-04
fix(deps): bump Pillow to 12.3.0

jamieQ:
## Summary

- bump the Pillow dep & min version to 12.3.0

Fixes EME-1277
Fixes EME-1278
Fixes EME-1279


pyproject.toml
```diff
@@ -35,7 +35,7 @@ dependencies = [
     "lief==0.17.2",
     "lzfse>=0.4.2",
     "objectstore-client>=0.1.0",
-    "pillow>=11.3.0",
+    "pillow>=12.3.0",
```
jamieQ, line 38: note, the lockfile value was 12.2.0, so this is not as big of a jump as the diff from this file suggests

chromy (approved):

pyproject.toml
chromy: Thanks 🙏🏻

NicoHinderling: nice catch, yea i was concerned about the 11->12 jump

NicoHinderling (approved):

PR #654, opened 2026-08-09
ref: record and collect some DEX processing stats

jamieQ:
Gather some additional telemetry for DEX processing to help investigate why some jobs take multiple minutes to compute class definitions. Current speculative theory is that we have superlinear scaling when searching through method/field annotation lists, which may become problematic for large apps.


chromy (approved):
lgtm, thanks! Interested to see the results.

PR #655, opened 2026-08-09
fix: Clean up bundletool temporary files

jamieQ:
Change the `build_apks()` implementation to clean up the intermediate dirs for APKs and keystores rather than "leaking" them.


src/launchpad/utils/android/bundletool.py
```diff
@@ -14,7 +14,6 @@
 
 from pydantic import BaseModel, Field
```
jamieQ, line 15: Note: recommend reviewing with whitespace changes hidden

chromy (approved):
thanks!

PR #656, opened 2026-08-12
perf(android): cache DEX lookup tables

jamieQ:
This change adds some caching into the DEX parsing logic in an attempt to improve performance for large apps. We now cache strings, type names, prototypes, and type lists after resolving them. Benchmarking locally against the hackernews app fixture showed a ~40% speed improvement, and somewhat reduced peak memory use.

### DIFF
--- a/src/launchpad/parsers/android/dex/dex_base_utils.py
+++ b/src/launchpad/parsers/android/dex/dex_base_utils.py
@@ -1,3 +1,7 @@
+import typing
+
+from dataclasses import dataclass, field
+
 from launchpad.parsers.android.dex.dex_field_parser import DexFieldParser
 from launchpad.parsers.android.dex.dex_method_parser import DexMethodParser
 from launchpad.parsers.android.dex.types import (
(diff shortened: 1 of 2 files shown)


src/launchpad/parsers/android/dex/dex_base_utils.py
```diff
@@ -316,41 +344,6 @@ def get_encoded_value(buffer_wrapper: BufferWrapper, header: DexFileHeader) -> E
             size=size,
         )
 
-    @staticmethod
-    def get_encoded_method_prototype(
```
jamieQ, line 320: tangential cleanup: this appeared to duplicate the `get_prototype()` impl, so just consolidated them.

src/launchpad/parsers/buffer_wrapper.py
```diff
@@ -39,6 +39,7 @@ def __init__(self, buffer: bytes, debug: bool = False) -> None:
         self.buffer = buffer
         self.cursor = 0
         self.debug = debug
+        self.cache: object | None = None
```
jamieQ, line 42: It's perhaps a bit unprincipled to stick the storage on this type rather than having one of the DEX parser types own it, but doing it this way has the perk of not requiring changing very much code, and associating a cache with each buffer.

rbro112 (approved):

PR #658, opened 2026-08-18
perf(android): target bundletool builds to device

jamieQ:
We now pass through the device spec `bundletool` option into the `build-apks` command, rather than just using it when invoking `extract-apks`. We do this unless we're producing a "universal" APK, since those options are mutually exclusive. Local testing showed this could cut down a meaningful amount of time for some large example cases (~43s -> ~33s), presumably by avoiding generating intermediate artifacts that we ultimately ignore.


tests/unit/utils/android/test_bundletool.py
```diff
@@ -0,0 +1,45 @@
+from pathlib import Path
```
chromy, line 1: Arguably this should live at:
- `tests/unit/utils/android/test_bundletool.py` rather than:
- `tests/unit/utils/test_bundletool.py` so as to match:
- `src/launchpad/utils/android/bundletool.py`

...but I'm not sure how closely we've followed that generally and the `src/launchpad/utils/android/` dir does not currently exist.

chromy (approved):
lgtm

tests/unit/utils/android/test_bundletool.py
jamieQ: That makes sense – will update it

PR #663, opened 2026-09-02
perf(size): Parallelize Apple binary analysis across processes

NicoHinderling:
The per-binary Mach-O analysis loop ran serially and dominated wall-clock time on a large anonymized iOS app (hundreds of binaries) — a major contributor to hitting the 15-minute deadline.

The work is CPU-bound Python (LIEF, symbol sizing, demangling), so this parallelizes it with a `ProcessPoolExecutor` to get past the GIL (threads were tried first and regressed). Workers parse from disk and results stay in input order, so `binary_analysis` and the treemap are unchanged. Worker count is a fixed `4` to match the pods' CPU request — prod sets no CPU limit, so `os.cpu_count()` would read node cores and oversubscribe — overridable via `LAUNCHPAD_BINARY_ANALYSIS_WORKERS`, where `0` runs in-process with no pool at all (the rollback switch). Demangling runs serial inside workers to cap subprocess fan-out, and the LIEF pre-parse cache is removed since workers couldn't use it anyway.

Workers start with a bare interpreter, so each configures its own logging and initializes its own Sentry client with the parent's config, continuing the parent's trace from propagated headers. Per-binary started/completed logs (with `elapsed_s`) fire inside the worker at the moment of the work and reach both stdout and Sentry Logs under the build's trace. Workers also return start/end timestamps and the parent rebuilds the per-binary span from them, so the trace waterfall shows real overlap.

Timeout safety was the main review concern, since a worker wedged in native LIEF ignores `SIGTERM`. **We traced how the taskbroker enforces the deadline**: it's an in-process `SIGALRM` raising `ProcessingDeadlineExceeded` inside the task process — there's no external kill. Because the main process is idle in `executor.map` when it fires, our `except` calls `executor.kill_workers()` (new in 3.14), which `SIGKILL`s the wedged workers directly (verified under `forkserver`, prod's start method). If the main process is hard-killed instead (e.g. OOM), it's the container's PID 1, so the runtime tears down the whole cgroup and takes the workers with it. **Either way, no orphaned binary-analysis subprocesses.**

Benchmark — large anonymized app, 8 workers on a 10-core dev box (shipped default is 4, so real-world gain is a bit smaller):

| Metric | Serial | This PR |
|---|---|---|
| Binary-analysis phase | ~255s | ~44s (~5.8×) |
| Total wall clock | ~461s | ~250s (~1.8×) |

Verified with in-process-vs-parallel equivalence, worker-count, worker-logging (stdout with request id, third-party suppression, and delivery to a local Sentry ingest stub under the parent's trace), and rebuilt-span tests.

### DIFF
--- a/src/launchpad/size/analyzers/apple.py
+++ b/src/launchpad/size/analyzers/apple.py
@@ -67,8 +67,6 @@ logger = get_logger(__name__)
 
 
 def _binary_worker_init() -> None:
-    # Each binary already runs in its own worker; keep per-binary demangling serial so
-    # the outer pool doesn't multiply into dozens of concurrent cwl-demangle subprocesses.
     os.environ["LAUNCHPAD_NO_PARALLEL_DEMANGLE"] = "true"
 
 
@@ -78,7 +76,6 @@ def _force_kill_pool(executor: ProcessPoolExecutor) -> None:
             proc.kill()
         except Exception:
             pass
-    # Best-effort cleanup: never raise, or we'd mask the exception being re-raised at the call site.
     try:
         executor.shutdown(wait=False, cancel_futures=True)
     except Exception:
@@ -86,8 +83,6 @@ def _force_kill_pool(executor: ProcessPoolExecutor) -> None:
 
 
 def _run_and_time(analyze: Any, binary_info: BinaryInfo) -> Tuple[Any, float]:
-    # Time each binary inside the worker that runs it, so per-binary duration survives even
-    # though the parent emits the completed logs in a batch after all results return.
     start = time.monotonic()
     return analyze(binary_info), time.monotonic() - start
 
@@ -190,8 +185,6 @@ class AppleAppAnalyzer:
 
             if workers > 1:
                 logger.debug(f"Analyzing binaries in parallel with {workers} processes")
-                # Workers re-parse from disk (lief_cache=None), so the parent's cached LIEF
-                # trees are dead weight; drop them so they don't stay resident all run.
                 artifact.get_lief_cache().clear()
                 analyze = partial(self._analyze_binary, app_bundle_path=app_bundle_path, lief_cache=None)
                 executor = ProcessPoolExecutor(max_workers=workers, initializer=_binary_worker_init)
@@ -494,8 +487,6 @@ class AppleAppAnalyzer:
                 configured = 0
             if configured >= 1:
                 return min(configured, num_binaries)
-        # Match the worker pod's whole-core CPU request; os.cpu_count() reads node cores
-        # (no CPU limit set), which would oversubscribe. Override via the env var above.
         return min(4, num_binaries)
 
     def _log_binary_started(self, binary_info: BinaryInfo, app_bundle_path: Path, extract_dir: Path) -> None:
--- a/src/launchpad/size/analyzers/apple.py
+++ b/src/launchpad/size/analyzers/apple.py
@@ -2,12 +2,17 @@
 
 from __future__ import annotations
 
+import ctypes
 import gc
 import os
+import signal
+import sys
 import tempfile
 import time
 
+from concurrent.futures import ProcessPoolExecutor
 from datetime import datetime
+from functools import partial
 from pathlib import Path
 from typing import Any, Dict, List, Tuple
 
@@ -62,6 +67,28 @@ from ..models.apple import (
 )
 
 logger = get_logger(__name__)
+
+
+def _binary_worker_init() -> None:
+    if not sys.platform.startswith("linux"):
+        return
+    try:
+        libc = ctypes.CDLL(None, use_errno=True)
+        libc.prctl.argtypes = [ctypes.c_int, ctypes.c_ulong, ctypes.c_ulong, ctypes.c_ulong, ctypes.c_ulong]
+        libc.prctl.restype = ctypes.c_int
+        PR_SET_PDEATHSIG = 1
+        libc.prctl(PR_SET_PDEATHSIG, signal.SIGKILL, 0, 0, 0)
+    except Exception:
+        logger.debug("Could not set PR_SET_PDEATHSIG for binary worker", exc_info=True)
+
+
+def _force_kill_pool(executor: ProcessPoolExecutor) -> None:
+    for proc in list(getattr(executor, "_processes", {}).values()):
+        try:
+            proc.kill()
+        except Exception:
+            pass
+    executor.shutdown(wait=False, cancel_futures=True)
 
 
 class AppleAppAnalyzer:
@@ -155,37 +182,33 @@ class AppleAppAnalyzer:
             binaries = artifact.get_all_binary_paths()
             logger.debug(f"Found {len(binaries)} binaries to analyze")
 
-            lief_cache = artifact.get_lief_cache()
+            extract_dir = artifact.get_extract_dir()
+            workers = self._binary_analysis_worker_count(len(binaries))
             for binary_info in binaries:
-                logger.info(
-                    "size.apple.binary_analysis_started",
-                    extra={
-                        "event": "size.binary_analysis_started",
-                        "binary_name": binary_info.name,
-                        "binary_path": str(binary_info.path.relative_to(app_bundle_path)),
-                        "has_dsym": binary_info.dsym_path is not None,
-                    },
-                )
-                if binary_info.dsym_path:
-                    logger.debug(
-                        f"Found dSYM file for {binary_info.name} at {binary_info.dsym_path.relative_to(artifact.get_extract_dir())}"
-                    )
-                binary = self._analyze_binary(binary_info, app_bundle_path, lief_cache)
+                self._log_binary_started(binary_info, app_bundle_path, extract_dir)
+
+            if workers > 1:
+                logger.debug(f"Analyzing binaries in parallel with {workers} processes")
+                analyze = partial(self._analyze_binary, app_bundle_path=app_bundle_path, lief_cache=None)
+                executor = ProcessPoolExecutor(max_workers=workers, initializer=_binary_worker_init)
+                try:
+                    results = list(executor.map(analyze, binaries))
+                    executor.shutdown(wait=True)
+                except BaseException:
+                    _force_kill_pool(executor)
+                    raise
+            else:
+                lief_cache = artifact.get_lief_cache()
+                results = []
+                for binary_info in binaries:
+                    results.append(self._analyze_binary(binary_info, app_bundle_path, lief_cache))
+                    gc.collect()
+
+            for binary_info, binary in zip(binaries, results):
                 if binary is not None:
                     binary_analysis.append(binary)
                     binary_analysis_map[str(binary_info.path.relative_to(app_bundle_path))] = binary
-
-                logger.info(
-                    "size.apple.binary_analysis_completed",
-                    extra={
-                        "event": "size.binary_analysis_completed",
-                        "binary_name": binary_info.name,
-                        "symbol_count": (len(binary.symbol_info.symbol_sizes) if binary.symbol_info else 0),
-                        "swift_types_count": (len(binary.symbol_info.swift_type_groups) if binary.symbol_info else 0),
-                        "objc_types_count": (len(binary.symbol_info.objc_type_groups) if binary.symbol_info else 0),
-                    },
-                )
-                gc.collect()
+                    self._log_binary_completed(binary_info, binary)
 
             hermes_reports = make_hermes_reports(app_bundle_path)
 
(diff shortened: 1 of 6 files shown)


NicoHinderling:
> _Resolved in [`f81d124`](https://github.com/getsentry/launchpad/commit/f81d12472f753ee7dbba37f4ebe7a592f57a393b)_
> 
> **Bug:** The `size.apple.binary_analysis_started` log events are now emitted in a single batch before analysis begins, making it impossible to measure per-binary analysis duration.Severity: LOW
> 
> Suggested Fix
> Move the `_log_binary_started` call back inside the main analysis loop, immediately before the `_analyze_binary` call for each binary. This will restore the original behavior of bracketing each binary's analysis with 'started' and 'completed' log events, allowing for accurate duration measurement.
> 
> Prompt for AI Agent
> ```
> Review the code at the location below. A potential bug has been identified by an AI
> agent. Verify if this is a real issue. If it is, propose a fix; if not, explain why it's
> not valid.
> 
> Location: src/launchpad/size/analyzers/apple.py#L190-L191
> 
> Potential issue: The refactoring introduced a new loop that emits all
> `size.apple.binary_analysis_started` log events before any binary analysis starts.
> Previously, the 'started' and 'completed' logs for each binary were emitted
> sequentially, bracketing the analysis work. With this change, all 'started' events have
> nearly identical timestamps, fired before the actual analysis loop begins. This makes it
> impossible to use these logs for per-binary progress tracking or to calculate the
> duration of individual binary analyses, which is a regression in observability.
> ```

i did a simple solution for this for now: "Time each binary's analysis inside the worker that runs it and return that duration with the result, so the parent's per-binary "completed" log carries an accurate elapsed_s even though the log lines are emitted together."

If we do land this PR, I'd plan to cut a followup PR to make sure that all the subprocess logs are piped back to the parent process for o11y convenience (ex. all trace info would stay intact)

src/launchpad/size/analyzers/apple.py
```diff
@@ -454,6 +476,45 @@ def _generate_insight_with_tracing(
             )
             return result
 
+    def _binary_analysis_worker_count(self, num_binaries: int) -> int:
+        if num_binaries <= 1:
+            return 1
+        override = os.getenv("LAUNCHPAD_BINARY_ANALYSIS_WORKERS")
```
trevor-e: nit: think this can be simplified if we combine `os.getenv("LAUNCHPAD_BINARY_ANALYSIS_WORKERS", "4")` as a default and then don't need the `if override` branch

src/launchpad/size/analyzers/apple.py
```diff
@@ -155,37 +178,36 @@ def analyze(self, artifact: AppleArtifact) -> AppleAnalysisResults:
...
-                    },
-                )
-                if binary_info.dsym_path:
-                    logger.debug(
-                        f"Found dSYM file for {binary_info.name} at {binary_info.dsym_path.relative_to(artifact.get_extract_dir())}"
-                    )
-                binary = self._analyze_binary(binary_info, app_bundle_path, lief_cache)
+                self._log_binary_started(binary_info, app_bundle_path, extract_dir)
+
+            if workers > 1:
```
trevor-e: Does this happen frequently? I imagine basically all apps at this point have more than one binary in them and we don't need to special-case this which will help simplify things a bit.

src/launchpad/size/analyzers/apple.py
```diff
@@ -155,37 +178,36 @@ def analyze(self, artifact: AppleArtifact) -> AppleAnalysisResults:
...
+                logger.debug(f"Analyzing binaries in parallel with {workers} processes")
+                artifact.get_lief_cache().clear()
```
trevor-e: Why do we have to clear the cache now?

src/launchpad/size/analyzers/apple.py
```diff
@@ -155,37 +178,36 @@ def analyze(self, artifact: AppleArtifact) -> AppleAnalysisResults:
...
+                analyze = partial(self._analyze_binary, app_bundle_path=app_bundle_path, lief_cache=None)
```
trevor-e: Actually is the cache needed at all then? I was wondering how that would work with a process pool but looks like we just always set `lief_cache=None` now.

src/launchpad/size/analyzers/apple.py
```diff
@@ -155,37 +178,36 @@ def analyze(self, artifact: AppleArtifact) -> AppleAnalysisResults:
...
-                        "binary_path": str(binary_info.path.relative_to(app_bundle_path)),
-                        "has_dsym": binary_info.dsym_path is not None,
-                    },
-                )
-                if binary_info.dsym_path:
-                    logger.debug(
-                        f"Found dSYM file for {binary_info.name} at {binary_info.dsym_path.relative_to(artifact.get_extract_dir())}"
-                    )
-                binary = self._analyze_binary(binary_info, app_bundle_path, lief_cache)
+                self._log_binary_started(binary_info, app_bundle_path, extract_dir)
```
trevor-e: This logging is kind of awkward now since this for loop is purely to log everything at once. Could we maybe push these logs into `_run_and_time` itself so they come in when the analyzing actually happens? Otherwise I could see it being confusing that they don't actually align with when the work actually happened.

src/launchpad/size/analyzers/apple.py
```diff
@@ -155,37 +178,36 @@ def analyze(self, artifact: AppleArtifact) -> AppleAnalysisResults:
...
-                    extra={
-                        "event": "size.binary_analysis_completed",
-                        "binary_name": binary_info.name,
-                        "symbol_count": (len(binary.symbol_info.symbol_sizes) if binary.symbol_info else 0),
-                        "swift_types_count": (len(binary.symbol_info.swift_type_groups) if binary.symbol_info else 0),
-                        "objc_types_count": (len(binary.symbol_info.objc_type_groups) if binary.symbol_info else 0),
-                    },
-                )
-                gc.collect()
+                    self._log_binary_completed(binary_info, binary, elapsed)
```
trevor-e: Same with this logging.

src/launchpad/size/analyzers/apple.py
```diff
@@ -64,6 +66,27 @@
 logger = get_logger(__name__)
 
 
+def _binary_worker_init() -> None:
+    os.environ["LAUNCHPAD_NO_PARALLEL_DEMANGLE"] = "true"
+
+
+def _force_kill_pool(executor: ProcessPoolExecutor) -> None:
+    for proc in list(getattr(executor, "_processes", {}).values()):
```
trevor-e: Python 3.14 has public methods for this now: `executor.kill_workers()` and `executor.terminate_workers()` depending on how graceful we want it to be.

src/launchpad/size/analyzers/apple.py
trevor-e: The more relevant part is whether we want to slow roll this out and keep the old logic side-by-side. I didn't see mention of a rollout plan in the description, are we going to hard cut over?

src/launchpad/size/analyzers/apple.py
```diff
@@ -155,37 +178,36 @@ def analyze(self, artifact: AppleArtifact) -> AppleAnalysisResults:
...
-                        f"Found dSYM file for {binary_info.name} at {binary_info.dsym_path.relative_to(artifact.get_extract_dir())}"
-                    )
-                binary = self._analyze_binary(binary_info, app_bundle_path, lief_cache)
+                self._log_binary_started(binary_info, app_bundle_path, extract_dir)
+
+            if workers > 1:
+                logger.debug(f"Analyzing binaries in parallel with {workers} processes")
+                artifact.get_lief_cache().clear()
+                analyze = partial(self._analyze_binary, app_bundle_path=app_bundle_path, lief_cache=None)
+                executor = ProcessPoolExecutor(max_workers=workers, initializer=_binary_worker_init)
```
trevor-e: Can we confirm the peak memory usage here too, I would need to dig further but if it forks the parent process it could get pretty high.

src/launchpad/size/analyzers/apple.py
NicoHinderling: Nice, didn't know that landed in 3.14. Switched to executor.kill_workers() 

Kept kill over terminate since the case we care about is a worker wedged in LIEF that ignores SIGTERM.

src/launchpad/size/analyzers/apple.py
NicoHinderling: get_all_binary_paths() parses every binary for dSYM UUID matching and caches the parsed FatBinary for dSYM-bearing ones. The serial path pops entries as it goes, but workers can't use the cache (LIEF objects don't pickle, forkserver doesn't inherit parent state), so nothing would ever free them. Clearing up front drops the parent's copies before 4 workers each build their own

NicoHinderling: actually nevermind, saw your other comment - good catch

removing the cache

src/launchpad/size/analyzers/apple.py
NicoHinderling: as for rollout, I was planning to merge but pause the pipeline so it only deploys to s4s2 and do some manual testing before fully rolling it out

> The more relevant part is whether we want to slow roll this out and keep the old logic side-by-side.

I don't plan on doing a slow rollout. I've tweaked the logic to keep the old path if `LAUNCHPAD_BINARY_ANALYSIS_WORKERS=0`

that way we can killswitch it if we need to by setting that env var to 0

if all goes well, i'll clean up the old path in a week or something

NicoHinderling: before kicking this back for you to review again though, ill investigate wiring up sentry-options into launchpad :loading:

src/launchpad/size/analyzers/apple.py
NicoHinderling: it doesn't fork the parent. agent test in the exact prod image that Python 3.14 on Linux defaults to forkserver, and nothing in launchpad overrides the start method. Forkserver workers are forked from a small helper process that starts fresh, not from the parent, so they inherit none of the parent's heap. I measured an idle worker at 15 MB earlier today. Each worker's memory is just the binary it's parsing.

src/launchpad/size/analyzers/apple.py
NicoHinderling: https://github.com/getsentry/launchpad/pull/668
https://github.com/getsentry/sentry-options-automator/pull/9596
https://github.com/getsentry/ops/pull/23267

i've posted and sent ken these PRs to enable sentry-options within launchpad. Assuming he reviews them before eod, I'll merge and then switch the code to use this instead of the env var for convenience

I wont merge until eod but for now the env var = 0 killswitch should be valid

src/launchpad/size/analyzers/apple.py
NicoHinderling: Measured peak memory properly this time: sum of RSS across the parent and every descendant, sampled every 0.5s, on the large app.

| Path | Peak tree RSS |
|---|---|
| In-process (`LAUNCHPAD_BINARY_ANALYSIS_WORKERS=0`) | 2.44 GB |
| 4 workers (default) | 2.53 GB |

~90 MB more, under 4%. The peak is dominated by parsing the single largest binary, which happens in one process either way; the other binaries in flight are small next to it.

src/launchpad/size/analyzers/apple.py
NicoHinderling: i was considering doing this as a follow up pr but ended up just doing the bigger change now

workers now get a QueueHandler and the parent drains the queue and re-emits through its own handlers with the request id attached. So the per-binary started/completed logs fire inside the worker at the moment of the work, and both parent-side loops are gone. 

Workers also return start/end timestamps and the parent rebuilds the per-binary span from them, so the trace waterfall shows real overlap.

src/launchpad/size/analyzers/apple.py
```diff
@@ -472,18 +590,13 @@ def _analyze_binary(
...
-            logger.debug(f"Binary not in LIEF cache, parsing now: {binary_path.name}")
-            with open(binary_path, "rb") as f:
-                config = lief.MachO.ParserConfig()
-                config.parse_dyld_exports = False
-                config.parse_dyld_bindings = False
-                config.parse_dyld_rebases = False
-
-                fat_binary = lief.MachO.parse(f, config)  # type: ignore
+        with open(binary_path, "rb") as f:
+            config = lief.MachO.ParserConfig()
```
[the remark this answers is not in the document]
NicoHinderling: not worrying abt that

src/launchpad/size/analyzers/apple.py
```diff
@@ -64,6 +73,62 @@
...
+def _binary_worker_init(log_queue: multiprocessing.Queue[logging.LogRecord], log_level: int) -> None:
+    os.environ["LAUNCHPAD_NO_PARALLEL_DEMANGLE"] = "true"
+    root = logging.getLogger()
+    root.handlers = [QueueHandler(log_queue)]
+    root.setLevel(log_level)
+
+
+class _TimedBinary(NamedTuple):
+    binary: MachOBinaryAnalysis | None
+    started_at: float
```
trevor-e: nit: is float right? i think what you had before with the monotomic timing was correct

src/launchpad/size/analyzers/apple.py
```diff
@@ -64,6 +73,62 @@
...
+        self._thread.start()
+        return self
+
+    def __exit__(self, *exc: object) -> None:
+        self._stop.set()
+        self._thread.join(timeout=2)
+        if self._thread.is_alive():
+            logger.warning("Worker log relay did not stop; a worker was likely killed mid-write")
+
+    def _drain(self) -> None:
```
trevor-e: I feel like there's got to be a much simpler way to do this but not sure the best practice. Like could we just pipe the stderr and stdout back to the main process and append the logs?

trevor-e (approved):

PR #664, opened 2026-09-02
perf(size): Dedupe images in ImageOptimization insight

NicoHinderling:
On a large anonymized iOS app, the image-optimization insight was ~99% of insights time — it re-encodes every optimizable image (PNG + JPEG + a HEIC probe), including asset-catalog images. Its 240s "timeout" also never fired when workers stalled, because the deadline was only checked between completions.

This dedupes images by content hash (encode each unique image once, then fan the result back out to every duplicate path with size/idiom/colorspace preserved, so treemap tagging is unchanged), enforces a real wall-clock budget via `as_completed(timeout=…)` + `shutdown(cancel_futures=True)`, and scales the worker pool to available CPUs.

Benchmark — large anonymized app. The binary phase is unchanged (256s vs 258s); insights only:

| Metric | Before | After |
|---|---|---|
| Insights phase | 139.8s | 118.0s (−16%) |
| image_opt output | 833 files, 31,296,600 B | identical |

Output is unchanged (the dedup fan-out preserves per-path reporting). The win is modest since most images are unique; the bigger value is the timeout fix removing a way to blow the budget. Gating the redundant HEIC probe is a further lever, deferred here since it would change reported savings. Tested with new dedup + wall-clock-timeout tests, plus `make check`.

### DIFF
--- a/src/launchpad/size/insights/apple/image_optimization.py
+++ b/src/launchpad/size/insights/apple/image_optimization.py
@@ -113,16 +113,21 @@ class BaseImageOptimizationInsight(Insight[ImageOptimizationInsightResult], ABC)
                 continue
             for f in group:
                 size_delta = f.size - rep_result.current_size
-                scaled = rep_result.model_copy(
-                    update={
-                        "file_path": f.path,
-                        "current_size": f.size,
-                        "minify_savings": max(0, rep_result.minify_savings + size_delta),
-                        "conversion_savings": max(0, rep_result.conversion_savings + size_delta),
-                        "idiom": f.idiom,
-                        "colorspace": f.colorspace,
-                    }
-                )
+                minify_savings = max(0, rep_result.minify_savings + size_delta)
+                conversion_savings = max(0, rep_result.conversion_savings + size_delta)
+                update: dict[str, object] = {
+                    "file_path": f.path,
+                    "current_size": f.size,
+                    "minify_savings": minify_savings,
+                    "conversion_savings": conversion_savings,
+                    "idiom": f.idiom,
+                    "colorspace": f.colorspace,
+                }
+                if rep_result.minified_size is not None:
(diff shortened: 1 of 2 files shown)


src/launchpad/size/insights/apple/image_optimization.py
```diff
@@ -75,31 +76,60 @@ def generate(self, input: InsightsInput) -> ImageOptimizationInsightResult | Non
...
+                self._TIMEOUT_SECONDS,
+            )
+        finally:
+            executor.shutdown(wait=False, cancel_futures=True)
+
+        results: List[OptimizableImageFile] = []
+        for content_hash, group in groups.items():
+            rep_result = analyzed.get(content_hash)
+            if rep_result is None:
+                continue
```
trevor-e, line 113: nit: seems like an error case to log

trevor-e (approved):

src/launchpad/size/insights/apple/image_optimization.py
NicoHinderling: Good instinct, but this one isn't actually an error path. `rep_result is None` here is dominated by the normal "image isn't worth optimizing" case — `_analyze_image_optimization` returns `None` whenever savings don't clear `MIN_SAVINGS_THRESHOLD`, which is the common outcome for most images. The genuine skip/error cases (no full path, empty, unidentifiable, exception) are each already logged inside `_analyze_image_optimization`, and the thread-pool exception + timeout cases are logged in the `as_completed` loop above. So logging here would fire on every non-optimizable image and mislabel a normal outcome as an error. Leaving as-is.

NicoHinderling: ( sorry for automated comment, responded without asking me - updating my markdown files to avoid this in the future -_- )

PR #665, opened 2026-09-02
feat(size): Log per-insight timing in Apple analyzer

NicoHinderling:
Insight generation logged nothing per-insight. On a large anonymized iOS app that timed out, `size.apple.generate_insights` was the last log before silence — a slow insight was indistinguishable from a hang.

This logs each insight's duration on completion: `size.apple.insight_completed` with `insight` and `elapsed_s`.

It's observability-only, so there's no material speedup and none measured (460.7s vs 439.8s, within noise). What it buys is attribution: it immediately isolated the culprit — `image_optimization` was 129.06s of 129.3s insights (99.8%), while the other 11 summed to ~0.27s.

Tested with a unit test asserting the completion log fires with the insight name + `elapsed_s`, plus `make check`.

### DIFF
--- a/src/launchpad/size/analyzers/apple.py
+++ b/src/launchpad/size/analyzers/apple.py
@@ -446,7 +446,13 @@ class AppleAppAnalyzer:
         self, insight_class: type, insights_input: InsightsInput, insight_name: str
     ) -> Any:
         with sentry_sdk.start_span(op="insight", description=f"apple.insights.{insight_name}"):
-            return insight_class().generate(insights_input)
+            started = time.monotonic()
+            result = insight_class().generate(insights_input)
+            logger.info(
+                "size.apple.insight_completed",
+                extra={"insight": insight_name, "elapsed_s": round(time.monotonic() - started, 3)},
+            )
+            return result
 
     @sentry_sdk.trace
     def _analyze_binary(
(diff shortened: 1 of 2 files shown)


src/launchpad/size/analyzers/apple.py
```diff
@@ -446,7 +446,13 @@ def _generate_insight_with_tracing(
         self, insight_class: type, insights_input: InsightsInput, insight_name: str
     ) -> Any:
         with sentry_sdk.start_span(op="insight", description=f"apple.insights.{insight_name}"):
-            return insight_class().generate(insights_input)
+            started = time.monotonic()
+            result = insight_class().generate(insights_input)
```
trevor-e, line 450: Am fine if we want to keep this for logging purposes but I'm a bit confused why we need it when we have span information right above? Can't we just look at the span timings in the trace to figure out the same thing?

trevor-e (approved):

src/launchpad/size/analyzers/apple.py
NicoHinderling: for a run that completes, the span indeed covers it. The log is for the timeout case: when a task is killed at the deadline mid-insights, the transaction is truncated and the running insight's span never closes, so the trace can't tell you which insight was grinding. 

The logs flush per-insight as each finishes, leaving a durable trail up to the kill, which was missing in the case of the customer's build.. so it seems nice to add just in case - at least as long as timeouts still occasionally happen in prod

PR #668, opened 2026-09-03
feat(options): Wire up sentry-options for dynamic config

NicoHinderling:
Launchpad has had no dynamic configuration system — every runtime knob is a process-start environment variable, so changing one means a redeploy. This wires up [sentry-options](https://github.com/getsentry/sentry-options) (Sentry's file/ConfigMap-backed options library) so launchpad can read values that are set live through sentry-options-automator, without shipping a new image.

The integration defines a `launchpad` namespace (`sentry-options/schemas/launchpad/schema.json`) with a single option to start, `projects.skip`. `get_service_config()` reads it and unions it with the existing `PROJECT_IDS_TO_SKIP` env var (deduped, order-preserving). The option defaults to an empty list, so this is a no-op until values are set in the automator — from then on the skip list becomes a dynamic control that no longer needs a redeploy. Env stays authoritative for whatever it already sets; the option is purely additive.

Reads are defensive by construction: a small `launchpad/options.py` helper caches an idempotent `init()` and every read falls back to the caller's default on any error, so a missing schema dir or a library hiccup can never break artifact processing. This matters because the actual read happens in the forkserver worker children; init runs lazily in whichever process reads first (plus explicit `init_options()` at the CLI group and `run_worker` startup for early validation). I verified the graceful-degradation path end-to-end — the CLI still runs cleanly even when the schema dir is absent.

Schemas are baked into the image at `/etc/sentry-options` and read via `SENTRY_OPTIONS_DIR` (set in the Dockerfile); tests and local dev point the same var at the in-repo `sentry-options/` directory. The package ships `py.typed`, so `ty` needs no extra config. Schema validity is enforced by `init()` during the test suite, so a malformed schema fails CI.

Part of a three-repo change. Two follow-ups make the option actually settable in production: a sentry-options-automator PR registering launchpad in `repos.json` (pinned to this branch) and adding `option-values/launchpad/`, and an ops PR adding the `options.sentry.io/{inject,namespace}` injector annotations to the launchpad taskworker deployment. This PR is safe to merge on its own — with no values set anywhere, behavior is unchanged.

Testing: added unit coverage for the options helper (default, override, fallback) and for the env/option union in `get_service_config`; full unit suite (362) passes, along with lint, format, and types.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -9,6 +9,8 @@ export SENTRY_BASE_URL="http://localhost:8000"
 export LAUNCHPAD_WORKER_RPC_HOST="localhost:50051"
 export LAUNCHPAD_WORKER_CONCURRENCY="16"
 
+export SENTRY_OPTIONS_DIR="${PWD}/sentry-options"
+
 # STATSD_HOST=... # defaults to 127.0.0.1
 # STATSD_PORT=... # defaults to 8125
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,7 +38,8 @@ ENV PYTHONUNBUFFERED=1 \
     PATH="/.venv/bin:$PATH" \
     UV_PROJECT_ENVIRONMENT=/.venv \
     UV_COMPILE_BYTECODE=1 \
-    UV_NO_CACHE=1
+    UV_NO_CACHE=1 \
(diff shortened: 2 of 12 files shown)


kenzoengineer (approved):
sentry options related changes lgtm!

PR #670, opened 2026-09-03
feat(logging): Stamp the artifact id onto every log record

NicoHinderling:
Stacked on #663.

Finding a build's logs today means locating the "Processing artifact <id> (project: ..., org: ...)" line, reading its `request_id`, and filtering on that. The artifact id itself is only in message text, never a structured field.

This extends `request_context()` in `tracing.py` to take `**fields`, stored in a contextvar alongside the request id and reset together with it. The existing `RequestLogFilter` copies whatever fields are set onto each record, the same way it already stamps `request_id`. `process_message` enters it with `artifact_id`, so every line for the rest of the task carries it. Project and org ids stay in the processing log line and on the Sentry scope tags, since the artifact id alone is enough to find a whole build. The binary-analysis worker relay from #663 captures the same fields at creation and stamps them onto relayed worker records, so the per-binary lines carry them too.

Since Sentry's logging integration forwards every non-standard record attribute as a log attribute, `artifact_id:758251` in the Sentry Logs search returns the whole build, parent and workers, and the same field is present in the JSON lines GCP and Datadog ingest.

Covered by unit tests for the contextvar and filter (stamping, scoping, coexistence with `request_id`), an assertion on the `process_message` log line, and the existing forkserver relay test now also checking `artifact_id` on worker records.

### DIFF
--- a/src/launchpad/tracing.py
+++ b/src/launchpad/tracing.py
@@ -4,6 +4,7 @@ from contextlib import contextmanager
 from contextvars import ContextVar
 
 _request_id: ContextVar[str | None] = ContextVar("request_id")
+_log_fields: ContextVar[dict[str, str]] = ContextVar("log_fields")
 
 
 @contextmanager
@@ -25,13 +26,32 @@ def bind_request_id(request_id: str) -> None:
     _request_id.set(request_id)
 
 
+@contextmanager
+def log_context(**fields: str):
+    """Attach structured fields (e.g. artifact_id) to every log record emitted within the block."""
+    token = _log_fields.set({**_log_fields.get({}), **fields})
+    try:
+        yield
+    finally:
+        _log_fields.reset(token)
+
+
+def current_log_fields() -> dict[str, str]:
+    return _log_fields.get({})
+
+
+def bind_log_fields(fields: dict[str, str]) -> None:
+    _log_fields.set(fields)
+
+
 class RequestLogFilter:
-    """Logging filter that adds request_id to log records.."""
+    """Logging filter that adds request_id and log_context fields to log records."""
 
     def filter(self, record) -> bool:
+        for key, value in _log_fields.get({}).items():
+            setattr(record, key, value)
         try:
-            request_id = _request_id.get()
-            record.request_id = request_id
+            record.request_id = _request_id.get()
         except LookupError:
             pass
         return True
--- a/src/launchpad/tracing.py
+++ b/src/launchpad/tracing.py
@@ -8,14 +8,15 @@ _log_fields: ContextVar[dict[str, str]] = ContextVar("log_fields")
 
 
 @contextmanager
-def request_context():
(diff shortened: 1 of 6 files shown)


src/launchpad/tracing.py
```diff
@@ -4,17 +4,19 @@
...
 _request_id: ContextVar[str | None] = ContextVar("request_id")
+_log_fields: ContextVar[dict[str, str]] = ContextVar("log_fields")
 
 
 @contextmanager
-def request_context():
-    """Create a request context with a unique request_id."""
-    request_id = str(uuid.uuid4())
-    token = _request_id.set(request_id)
+def request_context(**fields: str):
```
jamieQ: Do we want this to be fully generic vs just having it take an optional artifact id?

src/launchpad/tracing.py
```diff
@@ -25,13 +27,22 @@ def bind_request_id(request_id: str) -> None:
...
+    _log_fields.set(fields)
+
+
 class RequestLogFilter:
-    """Logging filter that adds request_id to log records.."""
+    """Logging filter that adds request_id and request_context fields to log records."""
 
     def filter(self, record) -> bool:
+        for key, value in _log_fields.get({}).items():
+            setattr(record, key, value)
```
jamieQ: Do we want to dynamically set everything here? Didn't we have a bug once by like trying to write to a reserved logging field?

src/launchpad/tracing.py
```diff
@@ -25,13 +27,22 @@ def bind_request_id(request_id: str) -> None:
     _request_id.set(request_id)
 
 
+def current_log_fields() -> dict[str, str]:
+    return _log_fields.get({})
+
+
+def bind_log_fields(fields: dict[str, str]) -> None:
+    _log_fields.set(fields)
```
jamieQ: Should we be making copies here (for the returned dict and the arg) so the contents can't be mutated without going through one of these?

tests/unit/test_tracing.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def _filtered_record() -> logging.LogRecord:
+    record = logging.LogRecord("t", logging.INFO, __file__, 1, "msg", None, None)
+    RequestLogFilter().filter(record)
+    return record
+
+
+def test_request_context_stamps_request_id_and_fields_onto_records() -> None:
+    with request_context(artifact_id="123", project_id="p"):
```
jamieQ: nit: do we care to test the nesting behavior of these context manager scopes (fields are overwritten IIUC?)?

jamieQ (approved):
Generally makes sense to me. Left some thoughts for your consideration.

src/launchpad/tracing.py
NicoHinderling: that was a relic of an iteration, good call

PR #679, opened 2026-09-13
perf(size): Reduce Apple symbol analysis memory

jamieQ:
Reduce memory used by Apple binary symbol analysis by adding slots to high-cardinality symbol records and reusing repeated section, segment, mangled-name, and module strings. Module reuse is thread-local, preserving parallel demangling without shared mutable state.

Production-shaped benchmarks reduced peak memory by 149–445 MB and process RSS by 226–785 MB across representative workloads.


src/launchpad/utils/apple/cwl_demangle.py
```diff
@@ -23,7 +24,14 @@
 _MAX_PARALLEL_DEMANGLE_WORKERS = 4
 
 
-@dataclass
+class _DemangleThreadState(threading.local):
+    def __init__(self) -> None:
+        # Maps each module-name value to the canonical string instance shared
+        # by results produced on this thread.
+        self.canonical_module_names: dict[str, str] = {}
```
trevor-e: I think this could be removed by instead using something like `sys.intern(module_name)` to share string instances.

trevor-e (approved):

src/launchpad/utils/apple/cwl_demangle.py
jamieQ: That sounds like a good idea that could simplify this a bit. Will test it out.

PR #695, opened 2026-09-23
fix(worker): Bump taskbroker-client to 0.20.26

evanh:
0.20.26 sends the broker hostname alone as the `:authority` instead of `host:port`, which is what the Cloud Service Mesh route matches on (getsentry/taskbroker#795). The s4s2 `task-launchpad-push-worker` pool can't be routed through envoy until launchpad is on it — getsentry/ops#23716 does that for getsentry, which is already on 0.20.26, and deliberately leaves launchpad out.

`BatchPushTaskWorker` folded into `PushTaskWorker` when batching became the default in getsentry/taskbroker#751, so `src/launchpad/worker/config.py` moves to `PushTaskWorker` and drops `update_in_batches=True`. Every other kwarg at that call site is still supported.

The lock also moves sentry-arroyo 2.39.2 → 2.44.1, sentry-sdk 2.61.1 → 2.70.0, msgpack 1.1.2 → 1.2.2 and adds prometheus-client, all forced by 0.20.26's own floors.


evanh:
```
$ uv run pytest tests -q --ignore=tests/e2e --ignore=tests/integration
3 failed, 369 passed in 39.39s

$ uv run pytest tests/unit/test_worker_config.py -q
9 passed
```

The three are pre-existing on `main` — `test_universal_apk`, `test_get_apksigner_certs` and `test_aggregate_symbols_mixed_modules` fail identically with the diff stashed and 0.20.6 reinstalled (they want android/swift tooling that isn't on this machine).

Collection failed before the `config.py` change with `ImportError: cannot import name 'BatchPushTaskWorker'`, which is how the rename surfaced.

```
$ uv run python -c "import importlib.metadata as m; print(m.version('taskbroker-client'))"
0.20.26
$ uv lock --check
Resolved 138 packages
```

Heads up on a likely conflict: your local `evanh/fix/cleanup-args` branch deletes `src/launchpad/worker/config.py`, which is the file this PR edits. Whichever lands second will need a hand.

enochtangg (approved):

